PR #7, opened 2020-08-19
Add 'set -u' to all shell scripts

DusanJovic-NOAA:
Only build scripts that are currently used by `stack custom.yaml` and `stack_wcoss_dell_p3.yaml` are tested.


aerorahul (approved):
in essence:
`value` of `${variable:-}` is `""` if `${variable}` is undefined.

PR #12, opened 2020-08-26
Add w3nco dependency to prod_util

kgerheiser:
prod_util depends on w3nco. I don't know why this failure hasn't shown up in other builds? 

https://github.com/NOAA-EMC/NCEPLIBS-prod_util/blob/b5674bb870bd336fa69bf6c35cf7c77a5dc2e462/CMakeLists.txt#L29

### DIFF
--- a/libs/build_nceplibs.sh
+++ b/libs/build_nceplibs.sh
@@ -97,6 +97,9 @@ if $MODULES; then
       module load ip
       module load sp
       ;;
+   prod_util)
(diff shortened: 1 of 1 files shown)


aerorahul (approved):
Good catch. 
I understand why it won't fail in CI because everything is installed in flat structure. 
This just failed on Hera for me.  We added the utils after the initial test on Hera and didn't get to test elsewhere.

PR #14, opened 2020-09-02
Update NCEPLIBS and enable OpenMP for SP

kgerheiser:
OpenMP is important for SP so enable it for all the HPC configs.

NCEPLIBS fixes for GCC 10, cleanup CMakeLists, and remove -xHOST from certain libraries

### DIFF
--- a/config/stack_custom.yaml
+++ b/config/stack_custom.yaml
@@ -90,86 +90,86 @@ tau2:
 
 bacio:
   build: YES
-  version: v2.4.0
-  install_as: 2.4.0
+  version: v2.4.1
+  install_as: 2.4.1
   openmp: OFF
 
 sigio:
   build: YES
-  version: v2.3.0
-  install_as: 2.3.0
+  version: v2.3.2
+  install_as: 2.3.2
   openmp: OFF
 
 sfcio:
   build: YES
-  version: v1.4.0
-  install_as: 1.4.0
+  version: v1.4.1
+  install_as: 1.4.1
   openmp: OFF
 
 gfsio:
   build: YES
-  version: v1.4.0
-  install_as: 1.4.0
+  version: v1.4.1
+  install_as: 1.4.1
   openmp: OFF
 
 w3nco:
   build: YES
-  version: v2.4.0
-  install_as: 2.4.0
+  version: v2.4.1
+  install_as: 2.4.1
   openmp: OFF
 
 sp:
   build: YES
-  version: v2.3.0
-  install_as: 2.3.0
+  version: v2.3.1
+  install_as: 2.3.1
   openmp: OFF
 
 ip:
   build: YES
-  version: v3.3.0
-  install_as: 3.3.0
+  version: v3.3.1
+  install_as: 3.3.1
   openmp: OFF
 
 ip2:
   build: NO
-  version: v1.1.0
-  install_as: 1.1.0
+  version: v1.1.1
+  install_as: 1.1.1
   openmp: OFF
 
 landsfcutil:
   build: YES
-  version: v2.4.0
-  install_as: 2.4.0
+  version: v2.4.1
+  install_as: 2.4.1
   openmp: OFF
 
 nemsio:
+  build: YES
+  version: v2.5.2
+  install_as: 2.5.2
+  openmp: OFF
+
+nemsiogfs:
   build: YES
   version: v2.5.1
   install_as: 2.5.1
   openmp: OFF
 
-nemsiogfs:
-  build: YES
-  version: v2.5.0
-  install_as: 2.5.0
-  openmp: OFF
-
 w3emc:
   build: YES
-  version: v2.7.0
-  install_as: 2.7.0
+  version: v2.7.1
+  install_as: 2.7.1
   openmp: OFF
 
 g2:
   build: YES
-  version: v3.4.0
-  install_as: 3.4.0
+  version: v3.4.1
+  install_as: 3.4.1
   openmp: OFF
 
 g2tmpl:
   build: YES
-  version: v1.9.0
-  install_as: 1.9.0
+  version: v1.9.1
+  install_as: 1.9.1
   openmp: OFF
 
 crtm:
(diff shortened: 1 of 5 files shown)


config/stack_custom.yaml
```diff
@@ -90,86 +90,86 @@ tau2:
...
+  install_as: 2.4.1
   openmp: OFF
 
 sp:
   build: YES
-  version: v2.3.0
-  install_as: 2.3.0
+  version: v2.3.1
+  install_as: 2.3.1
   openmp: OFF
```
aerorahul, line 125: reason to keep this `OFF`?

aerorahul (approved):
Thanks for updating the pointers to the NCEPLIBS.
Just one question re. keeping OPENMP OFF for sp in `stack_custom.yaml`.

Otherwise everything looks good.
Will merge after the actions complete.

config/stack_custom.yaml
kgerheiser: No

Is there a reason in general to disable OpenMP by default? I changed SP because I know that SP in particular makes heavy use of it to calculate transformations

aerorahul: No. There is no reason to turn it OFF.  It was the default in our NCEPLIBS, and I stuck with that in my iteration.
Just like I was building shared libs, but now we have turned them to static.

aerorahul:
@Hang-Lei-NOAA Why did you merge this? 
There is a discussion going on.
Also, the CI actions have not completed.

Hang-Lei-NOAA:
Okay, I reviewed the code. Agreed with it.
The only option is merge.

climbfuji:
> @Hang-Lei-NOAA Why did you merge this?
> There is a discussion going on.
> Also, the CI actions have not completed.

I don't know what you all discussed yesterday, but my feeling is that in general we shouldn't merge changes such as MPI OFF->ON without running a set of tests (most likely ufs-weather-model regression tests) that we previously agreed on. I see it stayed off, so that's good.

To @Hang-Lei-NOAA's defense, @aerorahul approved the PR. Maybe we shouldn't approve while there are ongoing discussions or while the CI tests are still running?

Hang-Lei-NOAA:
Sorry, i did not noticed this.
Rahual reminder me.
I have talked with Kyle to modify the changes.
He will push a fix on this based on discussion.

Hang

kgerheiser:
MPI or OpenMP? I did turn OpenMP on for SP in several configs.

climbfuji:
> MPI or OpenMP? I did turn OpenMP on for SP in several configs.

I am sorry, OpenMP of course. There are so many issues with linking to the correct thread-safe versions of runtime libraries in complex application, one never knows what happens unless you have a comprehensive set of tests that tell you that before-after is the same. The ufs-weather-model also has a threading test (and soon threading unit tests that run for more than just one config) which will allow us to catch any b4b differences when changing the number of threads (there shouldn't be any).

Hang-Lei-NOAA:
Agree,  Several nceplibs libraries are using openmp.
These have been tested with fv3GFS v16 for implementation.

climbfuji:
> Agree, Several nceplibs libraries are using openmp. These have been tested with fv3GFS v16 for implementation.
> […](#)

Ok, that's fine. Is this the case for SP? If so, no problem. Just throwing out some words of caution here. Once we use hpc-stack with the ufs-weather-model develop branch and other ufs codes, we will need to run a set of predefined regression tests before making any updates to the libraries.

kgerheiser:
I can either turn it back off or leave it on.

I'll make a new PR to enable it for others and we can test it before merging.

kgerheiser:
When I sent SP to NCO for implementation OpenMP is enabled

aerorahul:
> > Agree, Several nceplibs libraries are using openmp. These have been tested with fv3GFS v16 for implementation.
> > […](#)
> 
> Ok, that's fine. Is this the case for SP? If so, no problem. Just throwing out some words of caution here. Once we use hpc-stack with the ufs-weather-model develop branch and other ufs codes, we will need to run a set of predefined regression tests before making any updates to the libraries.

@climbfuji I think as we test these and other settings, things like this will settle down.  We rarely will change configurations radically. And even if we did, with a proper approach, we can have multiple stacks with different versions/combinations for development, regression testing and stable purposes.

As for code-review and CI, I view them as independent and complementary (and that depends on the reviewers); Some reviewers won't review until the CI finishes (which is fair).  I reviewed the code and approved it because the changes are very simple to review (I also expect CI to pass).  But, I would not merge the branch before the CI finishes. Just approval is not a sufficient condition to merge when CI is present.  E.g. last week was w3emc (I think).  You and I both approved the gcc-10 changes, but CI failed due to an `apt-get update` issue.

Hang-Lei-NOAA:
Agree with Rahul. I will be more careful.

climbfuji:
> > > Agree, Several nceplibs libraries are using openmp. These have been tested with fv3GFS v16 for implementation.
> > > […](#)
> > 
> > 
> > Ok, that's fine. Is this the case for SP? If so, no problem. Just throwing out some words of caution here. Once we use hpc-stack with the ufs-weather-model develop branch and other ufs codes, we will need to run a set of predefined regression tests before making any updates to the libraries.
> 
> @climbfuji I think as we test these and other settings, things like this will settle down. We rarely will change configurations radically. And even if we did, with a proper approach, we can have multiple stacks with different versions/combinations for development, regression testing and stable purposes.
> 
> As for code-review and CI, I view them as independent and complementary (and that depends on the reviewers); Some reviewers won't review until the CI finishes (which is fair). I reviewed the code and approved it because the changes are very simple to review (I also expect CI to pass). But, I would not merge the branch before the CI finishes. Just approval is not a sufficient condition to merge when CI is present. E.g. last week was w3emc (I think). You and I both approved the gcc-10 changes, but CI failed due to an `apt-get update` issue.

Ok, as long as we all know what to do and when then that's ok.

aerorahul:
Well, look what we have.
macOS action failed. Who wants to look?
@Hang-Lei-NOAA care to revert that merge?

Hang-Lei-NOAA:
Okay, I will revert it

aerorahul:
Screen Shot 2020-09-02 at 1 12 10 PM

Also fails on Orion.

PR #16, opened 2020-09-02
Remove OpenMP option from libraries that have no OpenMP option

kgerheiser:
Update build script to account for potentially empty OpenMP variable


aerorahul (approved):
looks good.
thanks

PR #17, opened 2020-09-03
Update NCEPLIBS

kgerheiser:
Fixes for OpenMP and MPI dependencies. MPI (for NetCDF) and OpenMP are now found automatically if needed.

As for enabling OpenMP, I can confirm that it has been enabled on all the NCEPLIBS that Hang and I installed on Orion, Hera, and WCOSS.

Left it off for Mac because Apple Clang doesn't have OpenMP.

A bunch of kinks have been worked out in NCEPLIBS and hopefully things will be more stable now.


aerorahul:
@kgerheiser 
```
Left it off for Mac because Apple Clang doesn't have OpenMP.

```

Is that true? I can build with OpenMP on my Mac with clang and gfortran-9.

Here is a grab from sp:
```
[ 31%] Building Fortran object src/CMakeFiles/sp_8.dir/sptrunv.f.o
cd /Users/rmahajan/scratch/NCEPLIBS/nceplibs-sp/build/src && /usr/local/bin/gfortran-9 -DAPPLE -DOPENMP -I/Users/rmahajan/scratch/NCEPLIBS/nceplibs-sp/build/src/include_8 -g -fbacktrace -fconvert=big-endian  -O3 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk -Jinclude_8 -fdefault-integer-8 -fdefault-real-8 -fopenmp -c /Users/rmahajan/scratch/NCEPLIBS/nceplibs-sp/src/sptrunv.f -o CMakeFiles/sp_8.dir/sptrunv.f.o
```
It adds the correct `-fopenmp` compiler option.

kgerheiser:
GCC  (gfortran) has OpenMP, but the Apple supplied clang (for C) doesn't. Maybe there's no C OpenMP, and in that case it's fine.

climbfuji:
> GCC (gfortran) has OpenMP, but the Apple supplied clang (for C) doesn't. Maybe there's no C OpenMP, and in that case it's fine.

Apple clang doesn't have OpenMP, that is correct. The homebrew LLVM clang has, but we always had problems linking the ufs-weather-model when OpenMP was enabled with LLVM clang. Therefore, OpenMP is set to off in ufs-weather-model when building with LLVM clang+ GNU gfortran.

climbfuji:
CI fails, error:
```
-- Found sigio: /home/runner/work/hpc-stack/hpc-stack/install-openmpi/lib/libsigio.a (found version "2.3.0")
28641
CMake Error at /usr/local/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:164 (message):
28642
  Could NOT find MPI (missing: MPI_C_FOUND C)
28643

28644
      Reason given by package: MPI component 'C' was requested, but language C is not enabled.  MPI component 'CXX' was requested, but language CXX is not enabled.  MPI component 'C' was requested, but language C is not enabled.  MPI component 'CXX' was requested, but language CXX is not enabled.  
28645

28646
Call Stack (most recent call first):
28647
  /usr/local/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:445 (_FPHSA_FAILURE_MESSAGE)
28648
  /usr/local/share/cmake-3.17/Modules/FindMPI.cmake:1717 (find_package_handle_standard_args)
28649
  CMakeModules/Modules/FindNetCDF.cmake:181 (find_package)
28650
  CMakeLists.txt:29 (find_package)
28651

28652

28653
-- Configuring incomplete, errors occurred!
28654
See also "/home/runner/work/hpc-stack/hpc-stack/openmpi/pkg/w3emc-v2.7.2/build/CMakeFiles/CMakeOutput.log".
28655
BUILD FAIL!  NCEPlib: w3emc Error:1
28656
##[error]Process completed with exit code 1.
```

kgerheiser:
Yes, I know the problem. 

To use FindMPI the language you're searching for must be enabled. When I tested FindNetCDF for MPI I used wgrib2, which enables C. W3emc doesn't so when you search for the C component of MPI it fails.

The solution is to search only for the MPI component that's enabled.

aerorahul:
> Yes, I know the problem.
> 
> To use FindMPI the language you're searching for must be enabled. When I tested FindNetCDF for MPI I used wgrib2, which enables C. W3emc doesn't so when you search for the C component of MPI it fails.
> 
> The solution is to search only for the MPI component that's enabled.

In that case, can we just remove it out of the for-loop and put in something like:
```
if(NetCDF_PARALLEL)
  find_package(MPI COMPONENTS ${_search_components})
endif()
```
right after it is determined that netcdf is built with parallel?

aerorahul:
> > Yes, I know the problem.
> > To use FindMPI the language you're searching for must be enabled. When I tested FindNetCDF for MPI I used wgrib2, which enables C. W3emc doesn't so when you search for the C component of MPI it fails.
> > The solution is to search only for the MPI component that's enabled.
> 
> In that case, can we just remove it out of the for-loop and put in something like:
> 
> ```
> if(NetCDF_PARALLEL)
>   find_package(MPI COMPONENTS ${_search_components})
> endif()
> ```
> 
> right after it is determined that netcdf is built with parallel?

A quick fix. See attached diff for `FindNetCDF.cmake`
[diff.txt](https://github.com/NOAA-EMC/hpc-stack/files/5171567/diff.txt)

kgerheiser:
That's basically what I came up with. ~~I added a `if(TARGET MPI::MPI_${_comp})`, but it's otherwise the same~~. I don't think that's necessary.

Anyway, I'm testing it out now.

kgerheiser:
That doesn't work. C is always in the search components.

I put it to `find_package(MPI)` with no COMPONENTS which causes it to search for all enabled languages.

kgerheiser:
Fixed. Let's see how these tests go.

kgerheiser:
I ran regression tests on Hera using these libraries with OpenMP enabled and it passed

aerorahul (approved):
alright. finally.

DusanJovic-NOAA (approved):

PR #18, opened 2020-09-11
Add VERSION file

kgerheiser:
This is read by setup_modules to create the correct hpc/x.y.z module


VERSION
```diff
@@ -1 +1 @@
-v1.0.0-beta1
\ No newline at end of file
+1.0.0-beta1
```
DusanJovic-NOAA: Add newline at the end of line/file

setup_modules.sh
```diff
@@ -122,13 +123,13 @@ fi
 # Deploy directory for stack modulefile
 $SUDO mkdir -p $PREFIX/modulefiles/stack/hpc
 $SUDO cp $HPC_STACK_ROOT/modulefiles/stack/hpc/hpc.lua \
-         $PREFIX/modulefiles/stack/hpc/1.0.0.lua
+         $PREFIX/modulefiles/stack/hpc//$HPC_STACK_VERSION.lua
```
aerorahul: why the extra leading`/` before `$HPC_STACK_VERSION`

setup_modules.sh
```diff
@@ -19,6 +19,7 @@ set -eu
 # root directory for the repository
 HPC_STACK_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
+HPC_STACK_VERSION="$(cat ${HPC_STACK_ROOT}/VERSION)"
```
aerorahul: use `head -n 1` instead of `cat`.  If for some reason, someone adds description or extra line or whatever, `cat` will give the entire contents of the file.

VERSION
kgerheiser: I fixed my editor

aerorahul (approved):
Thanks for versioning the stack.

Although, changing the version number changes nothing. This modulefile (`hpc.lua`) simply adds the stack path to MODULEPATH, and the stack path is the same (so far):  defined by `$PREFIX`. Unless `PREFIX` string contains this version in it, this has little effect.  If we do that, then every version will have to be built entirely, which breaks update mode for libraries.

Something to think about!

kgerheiser:
I think that's fine. I just wanted `hpc/x.y.z` to be consistent.

PR #22, opened 2020-09-17
Remove SUDO from hpc-stack.

aerorahul:
This PR:
- removes the `USE_SUDO` option from config files.
- removes `$SUDO` from all the build scripts.
- removes the `USE_SUDO` description from `README.md` file.

If installation into a system location is required, this can be achieved by executing `sudo` on the driver scripts `setup_modules.sh` or `build_stack.sh` themselves.


DusanJovic-NOAA (approved):

aerorahul:
The problem with this approach is `root` owns everything, including the build directory.

closing PR.

PR #24, opened 2020-09-24
Hierarchy of Docker containers

aerorahul:
This PR:
- adds Dockerfiles for a very basic set of Docker containers.
- A README to get started on how to create and use a container.

More information will be added as needed.

### DIFF
--- a/containers/Dockerfile.hpc.ubuntu.base
+++ b/containers/Dockerfile.hpc.ubuntu.base
@@ -1,0 +1,13 @@
+FROM ncep_base:ubuntu
+RUN git clone https://github.com/NOAA-EMC/hpc-stack && \
+    cd hpc-stack && export HPC_MPI=mpich/3.3.2 && \
+    ./build_stack.sh -p /home/builder/opt -c config/config_custom.sh -y config/stack_custom.yaml
+
+RUN echo 'export PATH=/home/builder/opt/bin:$PATH'      >> /etc/bash.bashrc && \
+    echo 'export LD_LIBRARY_PATH=/home/builder/opt/lib' >> /etc/bash.bashrc && \
+    echo 'export CMAKE_PREFIX_PATH=/home/builder/opt'   >> /etc/bash.bashrc && \
+    echo 'export FC=mpifort' >> /etc/bash.bashrc && \
+    echo 'export CC=mpicc'   >> /etc/bash.bashrc && \
+    echo 'export CXX=mpicxx' >> /etc/bash.bashrc
+
+CMD ["/bin/bash"]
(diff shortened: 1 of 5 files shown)


DusanJovic-NOAA (approved):

containers/Dockerfile.hpc.ubuntu.base
```diff
@@ -0,0 +1,13 @@
+FROM ncep_base:ubuntu
```
MinsukJi-NOAA, line 1: Where is ncep_base:ubuntu stored?

aerorahul: @MinsukJi-NOAA 
I was hoping to discuss that with you, @arunchawla-NOAA and whoever this is of interest to at EMC.
I think we can create a dockerhub account for noaa-emc and store these container images there.
If we had a cloud based storage such as AWS:S3 we could store the images there or may be even our FTP site.  I am not sure sure about the last one.
They can be pulled into github actions for CI or for development in a docker container on personal machines.
As we proceed, we can add scripts here to update the images on docker hub (or elsewhere) so it is more or less seamless.

I will add a set of another scripts for Singularity containers in a subsequent PR.

kgerheiser:
Just out of curiosity, what uses Intel's OneAPI?

kgerheiser (approved):

aerorahul:
> Just out of curiosity, what uses Intel's OneAPI?

I think the point is to use the base tools from OneAPI such as compilers, MPI and MKL that are part of the HPC toolkit.
Most of the OneAPI (for DSL etc) is useless to us.

Here is the [reference](https://software.intel.com/content/www/us/en/develop/download/oneapi-product-brief.html)

containers/Dockerfile.hpc.ubuntu.base
MinsukJi-NOAA: Yeah, I briefly looked into Dockerhub.  Individual accounts are free, but organizational accounts are not (https://www.docker.com/pricing).  We could create an "individual account" with name noaa-emc, but not sure if it's the best approach.
I agree that it would be beneficial to have an organizational storage place for docker (and other) images, wherever that may be. AWS is definitely an option. Stored images will be used for the CI, and maybe even for the community by allowing users to download images on a local computer and start running simulations.

MinsukJi-NOAA (approved):

arunchawla-NOAA:
We have approval from the big data project to store our fix files on AWS.
Dom is the POC. He is supposed to add a few people from EMC. I wanted to
unify the fix files (we now have multiple copies for SRW, MRW and S2S) and
then push them there. We can do the same for docker images.
---------------------------------------------------------------
Arun Chawla
Chief
Engineering & Implementation Branch
Room 2083
National Center for Weather & Climate Prediction
5830 University Research Court
College Park, MD 20740
Phone: 301-683-3740
Mobile: 240-564-5675
     Fax: 301-683-3703
------------------------------------------------------------

PR #25, opened 2020-09-26
add FMS to hpc-stack

aerorahul:
This PR:
- adds the capability to build FMS as part of the hpc-stack.
- It is default `OFF` for NOAA machines in `stack_noaa.yaml`.
- A modulefile is also created and FMS module can be loaded via `module load fms`.  The modulefile sets the environment variable `FMS_ROOT`

### DIFF
--- a/config/stack_custom.yaml
+++ b/config/stack_custom.yaml
@@ -76,6 +76,12 @@ esmf:
   enable_pnetcdf: NO
   debug: NO
 
+fms:
+  build: YES
+  version: master
+  repo: noaa-gfdl
+  cmake_opts: "-DGFS_PHYS=ON -DLARGEFILE=ON"
+
 bacio:
   build: YES
   version: v2.4.1
--- a/config/stack_custom.yaml
+++ b/config/stack_custom.yaml
@@ -77,7 +77,7 @@ esmf:
   debug: NO
 
 fms:
-  build: YES
+  build: NO
   version: master
   repo: noaa-gfdl
   cmake_opts: "-DGFS_PHYS=ON -DLARGEFILE=ON"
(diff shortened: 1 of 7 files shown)


config/stack_custom.yaml
```diff
@@ -76,6 +76,12 @@ esmf:
   enable_pnetcdf: NO
   debug: NO
 
+fms:
+  build: NO
+  version: master
```
DusanJovic-NOAA, line 81: I think we should explicitly specify tag or version number.

aerorahul: Indeed. But since currently there is no official "tag" with the cmake build, master is all we have.

Also, the github action failed because of a `gettid` issue in Ubuntu in `affinity.c`.  I know this is an issue with the newer `glibc` installations.  Would @DusanJovic-NOAA or @kgerheiser have a solution for this action?  I would ideally, like to keep this `build: YES` in the action, even if it is `OFF` in `stack_noaa.yaml` for now.

aerorahul: Are there any more comments?
Since this is OFF by default, does it matter at the present time?

DusanJovic-NOAA (approved):

config/stack_custom.yaml
kgerheiser: I feel like I've seen that affinity.c bug before, though I don't exactly rememeber

kgerheiser (approved):

config/stack_custom.yaml
aerorahul: The part of the code is wrapped under an `#ifndef HAVE_GETTID`
```
#ifndef HAVE_GETTID
static pid_t gettid(void)
{
#if defined(__APPLE__)
  uint64_t tid64;
  pthread_threadid_np(NULL, &tid64);
  pid_t tid = (pid_t)tid64;
#else
  pid_t tid = syscall(__NR_gettid);
#endif
  return tid;
}
#endif
```

Other than GFDL fixing this, we could try defining `HAVE_GETTID` for the ubuntu github action.
I can test this some other time.

PR #26, opened 2020-10-05
Add quotes to executable names in prod_util module

kgerheiser:
Otherwise they're interpreted as empty variables


aerorahul (approved):
yep. my bad.

PR #30, opened 2020-10-20
Update Jasper to fix 'dangerous use of mktemp'

kgerheiser:
Jasper was updated some security fixes, including one for this.

Tested with NCEPLIBS and the weather model

Fixes #30

### DIFF
--- a/config/stack_custom.yaml
+++ b/config/stack_custom.yaml
@@ -29,7 +29,7 @@ szip:
 
 jasper:
   build: YES
-  version: 2.0.15
+  version: 2.0.22
 
 udunits:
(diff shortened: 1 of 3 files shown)


DusanJovic-NOAA (approved):

aerorahul (approved):

kgerheiser:
Weird failures related to uploading logs it looks like

aerorahul:
> Weird failures related to uploading logs it looks like

I have seen that, sometimes times out in the action.  Not sure how to fix that other that re-run.
I am going to merge this and let the merge into debug action run this again.

PR #38, opened 2020-10-26
Build static libjpeg if requested, use hpc-stack libjpeg if built in jasper and NCEPLIBS; add NCAR Cheyenne

climbfuji:
This PR fixes a few issues with libjpeg that surfaced when trying to install libjpeg on Cheyenne, see https://github.com/NOAA-EMC/hpc-stack/issues/39.

It also adds the configuration files for Cheyenne with Intel and GNU, and a separate NCAR YAML stack config. The latter differs from the NOAA config in building a static version of libjpeg.

Limitations:
- I fixed the jasper cmake build only, i.e. only for newer versions of jasper.
- The ufs-weather-model executable now launches, i.e. the library load problem is solved, but terminates immediately after the ESMF init. It also prints the ESMF banner ntasks number of times. I have seen this before, presumably ESMF is not built correctly on SGI MPI systems. I can fix this in this or a follow-up PR, whatever is better.

### DIFF
--- a/config/config_cheyenne_gnu.sh
+++ b/config/config_cheyenne_gnu.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Compiler/MPI combination
+export HPC_COMPILER="gnu/9.1.0"
+export HPC_MPI="mpt/2.22"
+
+# Build options
+export USE_SUDO=N
+export PKGDIR=pkg
+export LOGDIR=log
+export OVERWRITE=Y
+export NTHREADS=8
+export   MAKE_CHECK=N
+export MAKE_VERBOSE=N
+export   MAKE_CLEAN=N
+export DOWNLOAD_ONLY=N
+export STACK_EXIT_ON_FAIL=Y
+export WGET="wget -nv"
+
+# Load these basic modules for Hera
+module load cmake/3.16.4
--- a/config/config_cheyenne_gnu.sh
+++ b/config/config_cheyenne_gnu.sh
@@ -17,5 +17,5 @@ export DOWNLOAD_ONLY=N
 export STACK_EXIT_ON_FAIL=Y
 export WGET="wget -nv"
 
-# Load these basic modules for Hera
+# Load these basic modules for Cheyenne
 module load cmake/3.16.4
--- a/config/config_cheyenne_intel.sh
+++ b/config/config_cheyenne_intel.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Compiler/MPI combination
+export HPC_COMPILER="intel/19.1.1"
+export HPC_MPI="mpt/2.22"
+
+# Build options
+export USE_SUDO=N
+export PKGDIR=pkg
+export LOGDIR=log
+export OVERWRITE=Y
+export NTHREADS=8
+export   MAKE_CHECK=N
+export MAKE_VERBOSE=N
+export   MAKE_CLEAN=N
+export DOWNLOAD_ONLY=N
+export STACK_EXIT_ON_FAIL=Y
+export WGET="wget -nv"
+
+# Load these basic modules for Hera
+module load cmake/3.16.4
--- a/config/config_cheyenne_intel.sh
+++ b/config/config_cheyenne_intel.sh
@@ -17,5 +17,5 @@ export DOWNLOAD_ONLY=N
 export STACK_EXIT_ON_FAIL=Y
 export WGET="wget -nv"
 
-# Load these basic modules for Hera
+# Load these basic modules for Cheyenne
 module load cmake/3.16.4
--- a/config/stack_ncar.yaml
+++ b/config/stack_ncar.yaml
@@ -1,0 +1,243 @@
+cmake:
+  build: NO
+  version: 3.17.2
+
+jpeg:
+  build: YES
+  shared: NO
+  version: 9.1.0
+
+zlib:
+  build: YES
+  shared: NO
+  version: 1.2.11
+
+png:
+  build: YES
+  shared: NO
+  version: 1.6.35
+
+szip:
+  build: YES
+  shared: NO
+  version: 2.1.1
+
+jasper:
+  build: YES
+  shared: NO
+  version: 2.0.22
+
+udunits:
+  build: YES
+  shared: NO
+  version: 2.2.26
+
+hdf5:
+  build: YES
+  version: 1.10.6
+  shared: NO
+  enable_szip: NO
+  enable_zlib: YES
+
+pnetcdf:
+  build: NO
+  version: 1.12.1
+  shared: NO
+
+netcdf:
+  build: YES
+  shared: NO
+  enable_pnetcdf: NO
+  version_c: 4.7.4
+  version_f: 4.5.3
+  version_cxx: 4.3.1
+
+nccmp:
+  build: YES
+  version: 1.8.7.0
+
+nco:
+  build: NO
+  version: 4.9.3
+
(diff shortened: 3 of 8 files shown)


libs/build_jasper.sh
```diff
@@ -56,13 +56,18 @@ else
     useCmake=NO
 fi
 
+if [[ "${STACK_jpeg_build}" == "YES" ]]; then
+    module load jpeg
```
aerorahul: Would just a `module try-load jpeg` work? Because if one installed jpeg previously and now just updating jasper, the if block would return false.

libs/build_jpeg.sh
```diff
@@ -40,11 +42,14 @@ sourceDir=$PWD
...
 
+[[ $enable_shared =~ [yYtT] ]] && shared_flags="" || shared_flags="-DBUILD_STATIC=ON"
+
 cmake $sourceDir \
   -DCMAKE_INSTALL_PREFIX=$prefix \
   -DBUILD_TESTS=ON \
   -DBUILD_EXECUTABLES=ON \
-  -DCMAKE_BUILD_TYPE=RELEASE
+  -DCMAKE_BUILD_TYPE=RELEASE \
+  ${shared_flags}
```
aerorahul: This is fine. 
Can you see `bugfix/jpeg` and please add the `check_flags` logic from that branch.  It will save us another PR.

aerorahul (commented):
Looks good.
Just a couple minor comments.

Also, can you please add `shared: NO` in `stack_noaa.yaml` section for `jpeg`.

See `bugfix/jpeg` branch.  it will save us another PR.

Thanks.

libs/build_jpeg.sh
climbfuji: Should be done now, please check if I missed something.

libs/build_jasper.sh
climbfuji: Yes, I came across that only later when I looked at the NCEPLIBS scripts. Should be fixed now, please check.

config/config_cheyenne_gnu.sh
```diff
@@ -0,0 +1,21 @@
...
+export OVERWRITE=Y
+export NTHREADS=8
+export   MAKE_CHECK=N
+export MAKE_VERBOSE=N
+export   MAKE_CLEAN=N
+export DOWNLOAD_ONLY=N
+export STACK_EXIT_ON_FAIL=Y
+export WGET="wget -nv"
+
+# Load these basic modules for Hera
```
aerorahul: you mean Cheyenne

config/config_cheyenne_intel.sh
```diff
@@ -0,0 +1,21 @@
...
+export OVERWRITE=Y
+export NTHREADS=8
+export   MAKE_CHECK=N
+export MAKE_VERBOSE=N
+export   MAKE_CLEAN=N
+export DOWNLOAD_ONLY=N
+export STACK_EXIT_ON_FAIL=Y
+export WGET="wget -nv"
+
+# Load these basic modules for Hera
```
aerorahul: Cheyenne here too.

config/stack_ncar.yaml
```diff
@@ -0,0 +1,243 @@
+cmake:
```
aerorahul, line 1: how different is `stack_ncar.yaml` to `stack_noaa.yaml`?
Just curious.

config/config_cheyenne_gnu.sh
climbfuji: fixed for both intel and gnu

config/stack_noaa.yaml
```diff
@@ -4,6 +4,7 @@ cmake:
 
 jpeg:
   build: NO
+  shared: NO
```
aerorahul, line 7: If `build: NO` is the only difference between `stack_noaa.yaml` and `stack_ncar.yaml`, do you see any harm in turning jpeg to YES in `stack_noaa.yaml`?
I don't. May be @kgerheiser has some thoughts.

libs/build_jasper.sh
```diff
@@ -56,13 +56,18 @@ else
     useCmake=NO
 fi
 
+# Load jpeg module if created by hpc-stack; requires setting
+# MAKE_POLICY_DEFAULT_CMP0074 to new below so that JPEG_ROOT is searched
+module try-load jpeg
```
aerorahul: This has to be up top in the `if $MODULES` block.

aerorahul (commented):
A few more comments.

config/stack_ncar.yaml
climbfuji: Just the build-jpeg YES|NO, correct.

config/stack_noaa.yaml
climbfuji: I would prefer building it per default on NOAA platforms, too. In that case we could ditch the stack_ncar.yaml file again.

libs/build_jasper.sh
climbfuji: It hopefully is now.

config/stack_noaa.yaml
aerorahul: I would in that case suggest just build jpeg on NOAA platforms and ditch `stack_ncar.yaml`

aerorahul (approved):
All good.  
Thanks for the fixes and porting to Cheyenne.
Look at the comment on jpeg for noaa.  We can build jpeg on noaa too and ditch stack_ncar.yaml.
I am fine with having both too.

climbfuji:
> All good.
> Thanks for the fixes and porting to Cheyenne.
> Look at the comment on jpeg for noaa. We can build jpeg on noaa too and ditch stack_ncar.yaml.
> I am fine with having both too.

Ok, thanks. I'll wait if @kgerheiser has anything to add. I am also trying a fix for the ESMF build on Cheyenne right now.

climbfuji:
I just pushed the esmf configuration update for SGI MPT (used on Cheyenne). Works just fine. After deciding what to do with building jpeg on NOAA platforms or not (and potentially removing the NCAR config), this PR is ready to merge.

config/config_cheyenne_gnu.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
aerorahul, line 1: One last comment otherwise happy to merge.
Do we need `config_cheyenne_gnu.sh` and `config_cheyenne_intel.sh`
Can we just have one `config_cheyenne.sh`
See `config_hera.sh` for example.

climbfuji: I remember we "slacked" about this before. I want to minimize the amount of work and reduce the potential of making mistakes when rolling out a new version of the stack. All it should take is a no-brainer exercise of checking out the code, running the seup_modules script and the build_stack script. No editing of whatsoever files. This should be a failsafe exercise at 3am in the morning when somebody calls you that the libraries got deleted accidentally and they need them back right now. We should do the same for every other platform for which we support multiple compilers, in my opinion. Right now only hera, i.e. `config_hera_gnu.sh` and `config_hera_intel.sh`. And possibly append `_intel` for all the other platforms.

aerorahul: sold!

climbfuji: Thanks, appreciate it. All CI tests passed, let's wait to hear from @kgerheiser about the jpeg build on NOAA platforms.

climbfuji: or maybe not ;-)

PR #63, opened 2020-11-14
add UPP to hpc-stack

### LINKED ISSUE #31: Update the nceppost with request by using the 10.0.0 version
Wen has added the nceppost with UPP version of 10.0.0. She request the installation test through hpc-stack.
We will first try it from the EMC_POST/develop branch, on each machine.
If it is okay, she needs to create a release to formally included in the hpc-stack.

A branch has been created to this issue
feature/postupdate

aerorahul:
This PR:
- adds UPP v10 to the hpc-stack
- retains nceppost until it is ready to be removed and replaced by UPP for post library in UFS
- fixes #31 

Change-Id: I09a155ef0e026d8fac682390f05e9c2dd9c7c2d4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 *.tgz
 *.bz2
 *.log*
+nohup*
 
 # Ignore the contents of these directories
 pkg*/
--- a/build_stack.sh
+++ b/build_stack.sh
@@ -151,7 +151,7 @@ build_lib pio
 
 # UFS 3rd party dependencies
 
-#build_lib esmf
+build_lib esmf
 build_lib fms
 
 # NCEPlibs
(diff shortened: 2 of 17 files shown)


aerorahul:
Looks like there is a bug in upp_v10.0.0

kgerheiser:
I don't know why it does that here specifically. I was able to build the tag fine, and it builds on the Action I setup for their repository.

aerorahul:
Yeah. I haven't got to why yet.  Help appreciated.  Mind boggling indeed.

climbfuji:
What is the problem?

kgerheiser:
https://github.com/NOAA-EMC/hpc-stack/pull/63/checks#step:4:23250

edwardhartnett:
Do we need this PR to work for the 1.1.0 release? Seems like that is implied by #71

aerorahul:
Yes.

kgerheiser:
Maybe related, maybe not, but is there a problem with building two versions of UPP, v10.0.0 and the dcea26 version? When installed in a flat structure, as we do on the Github Runner, modules like `vrbls3d` will overwrite one another? Or it picks up the one module first and it doesn't contain some of the variables?

aerorahul:
Good point @kgerheiser
I'll disable nceppost in the runner and keep UPP and then re-run the ci. 
Let's see.

climbfuji:
> Good point @kgerheiser
> I'll disable nceppost in the runner and keep UPP and then re-run the ci.
> Let's see.

Great finding, @kgerheiser. We don't need both at the same time, as far as I can see.

aerorahul:
reopened this PR.
The CI was successful with nceppost turned off.  @kgerheiser was correct to identify the issue in the comments above.  Thanks @kgerheiser

aerorahul:
woops. missing something here.

PR #66, opened 2020-11-16
prevent over-writing already created metamodules and stack modulefile

### LINKED ISSUE #65: Improve setup_modules.sh interface
Currently `setup_modules.sh` does not check if the meta modulefiles for compiler and mpi as well as for the stack are present or not.

Add a layer of security that warns the user that modulefiles for the compiler and MPI are already present and queries the user if they want to over-write them.  In most (almost all), the answer is NO, since these just load the underlying compiler module file or use the system compiler/MPI.

aerorahul:
This PR:
- adds a layer of security that warns the user that modulefiles for the compiler and MPI are already present and queries the user if they want to over-write them.  In most (almost all), the answer is NO, since these just load the underlying compiler module file or use the system compiler/MPI.
- fixes #65


kgerheiser:
Worked for me

kgerheiser (approved):

PR #68, opened 2020-11-16
Add stack yaml file for ufs weather ci test

MinsukJi-NOAA:
Do not build JEDI related libraries to save space when running CI test for ufs-weather-model.


aerorahul:
@MinsukJi-NOAA 
You can just delete the sections that you don't want to build e.g. everything after `boost`, just remove them from the yaml.

Also, do you really want to build shared libs for netcdf, hdf5, etc?

MinsukJi-NOAA:
> @MinsukJi-NOAA
> You can just delete the sections that you don't want to build e.g. everything after `boost`, just remove them from the yaml.
> 
> Also, do you really want to build shared libs for netcdf, hdf5, etc?

I will delete them.

I didn't have problem building and running the ufs-weahter-model.  What are the benefits you see for using static library?

aerorahul:
There is no problem at all.  Just consistent with how the stack is built on NOAA HPC's.

aerorahul (approved):
looks good.

PR #70, opened 2020-11-16
Bring the stack to Gaea.

### LINKED ISSUE #34: Question: installing hpc-stack on gaea
I will need to install hpc-stack on gaea.

All modules generated by hpc-stack are lua modules, which gaea doesn't understand. gaea knows tcl modules, and so does every other system.

What is the strategy for supporting gaea (after all, it is a NOAA RDHPC system)?

The quick-and-dirty `lua2tcl.py` script I added to NCEPLIBS won't work out of the box for the hpc-stack modules I guess, and it's really only duct tape instead of doing it right.

aerorahul:
This PR brings hpc-stack to Gaea. 
- add a config and yaml file for Gaea.
- Gaea does not use lmod.  @climbfuji built lmod and is/can be used if one wishes to use LMod on Gaea.  Building without modules is also allowed.

Using lmod on Gaea needs a couple of scripts from @climbfuji to set up lmod correctly.  The script is sourced in `config_gaea.sh`.

The UFS weather model and UFS utils have been built with this stack and ran on Gaea.
This has been verified by @climbfuji 

There is ongoing discussion in #34

fixes #34 

Contributors: @climbfuji


config/config_gaea.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# Compiler/MPI combination
+export HPC_COMPILER="intel/18.0.6.288"
+export HPC_MPI="impi/2018.0.4"
```
climbfuji: This line needs to be replaced with
```
export HPC_MPI="cray-mpich/7.7.11"
```
Loading impi probably has no effect, due to the Cray compiler wrappers being used as CC,FC,CXX (serial, MPI) below, but we need to make sure we load the correct cray mpich.

climbfuji (commented):
I am still working on the gaea build using feature/hpc-stack-nco and these changes. But I am certain about the one change impi->cray-mpich.

aerorahul:
That export line (and the compiler one) has no impact whatsoever because we load the modules explicitly in the config.

config/config_gaea.sh
aerorahul: Yeah. Definitely need to load the cray-mpich module correctly

climbfuji:
> That export line (and the compiler one) has no impact whatsoever because we load the modules explicitly in the config.

But shouldn't we try to be as consistent as possible? This is confusing at least, in particular for people who don't know what hpc-stack is doing under the hood.

aerorahul:
Yes. Definitely. 
The comment was for your understanding.

config/config_gaea.sh
aerorahul: done in 1dc96f0

aerorahul:
Can I get a review for this?

climbfuji:
> Can I get a review for this?

This needs update to work with lua modules, and fixes to `build_esmf.sh` after making the switch from impi to cray-mpich. I am working on this today, will send you a PR to incorporate in your branch in a few hours.

climbfuji (approved):
I used this version to compile hpc-stack on gaea, compile ufs-weather-model hash 2e25df5fe952d27355ed58963148f46b82565469 (i.e. before updating the ufs-weather-model to use the feature/hpc-stack-nco branch, for which I have a separate PR for gaea, https://github.com/NOAA-EMC/hpc-stack/pull/74), and to run the model using `rt.sh`.

config/stack_gaea.yaml
```diff
@@ -147,6 +147,11 @@ g2:
   version: v3.4.1
   install_as: 3.4.1
 
+g2c:
+  build: YES
```
climbfuji: Change this to NO please and we are ready to merge. Thanks!

climbfuji: Done - let's someone merge it.

aerorahul: done.

kgerheiser (approved):

PR #76, opened 2020-11-18
Update wgrib2 with fixes for CMake build

### LINKED ISSUE #75: Bugfix for wgrib2
wgrib2's build had a bug that caused compile definitions to not be passed correctly and resulted in broken features

Update wgrib2 to the latest release

kgerheiser:
Update wgrib2 to `v2.0.8-cmake-v6` with fixes in the CMake build

Also re-enable building wgrib2 with USE_IPOLATES enabled

Fixes #75 and #80

### DIFF
--- a/config/stack_mac.yaml
+++ b/config/stack_mac.yaml
@@ -120,7 +120,7 @@ ip:
   openmp: ON
 
 ip2:
-  build: NO
+  build: YES
   version: v1.1.2
   install_as: 1.1.2
   openmp: ON
@@ -190,11 +190,11 @@ bufr:
 
 wgrib2:
   build: YES
-  version: v2.0.8-cmake-v5
+  version: v2.0.8-cmake-v6
   install_as: 2.0.8
   openmp: OFF
-  spectral: OFF
-  ipolates: 0
+  spectral: ON
+  ipolates: 3
 
 prod_util:
   build: YES
--- a/config/stack_custom.yaml
+++ b/config/stack_custom.yaml
@@ -120,7 +120,7 @@ ip:
   openmp: ON
 
 ip2:
-  build: NO
+  build: YES
   version: v1.1.2
   install_as: 1.1.2
   openmp: ON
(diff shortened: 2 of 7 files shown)


kgerheiser:
Ah, there's some leftover submodule that hpc-stack doesn't like.

config/stack_mac.yaml
```diff
@@ -169,13 +169,13 @@ nceppost:
...
   install_as: dceca26
-  openmp: ON
+  openmp: OFF
 
 wrf_io:
   build: YES
   version: v1.1.1-cmake-v5
   install_as: 1.1.1
-  openmp: ON
+  openmp: OFF
```
edwardhartnett: Why do we turn off OpenMP here?

kgerheiser: No, real reason for wrf_io or post. I'll selectively re-enable it once I get this to build properly.

Clang on macOS doesn't support OpenMP, so it should be disabled for wgrib2 (C code). But somehow OpenMP_C was being propagated to wgrib2 because ip2 or sp are using it resulting in a build failure. OpenMP is marked as found through the CMake config for ip2 when wgrib2 searches for it (it runs find_package(OpenMP) for Fortran), so wgrib2 thinks OpenMP is available (it is for Fortran, but not C), and it tries to link to it and it's not available.

edwardhartnett: OK, thanks for the explanation!

climbfuji (changes requested):
I don't know the details behind version `v2.0.8-cmake-v6` of wgrib2, but these changes look good to me. I am missing the changes in `stack_gaea.yaml`, though. Did you update your branch with the latest changes from develop and check that all existing `*.yaml` files are updated?

kgerheiser:
No, I forgot there were new stack files. Will update.

aerorahul:
@kgerheiser 
Is there no clash with the fortran modules created by ip and ip2 in the flat structure?
Is that resolved in ip2?

kgerheiser:
It was fixed

edwardhartnett (approved):

kgerheiser:
@climbfuji could you re-review this? It's blocked until you do.

climbfuji (approved):

Issue #41, opened 2020-11-03
install esmf810bs36 in hpc stack


junwang-noaa:
ESMF team has updated esmf to esmf810bs36, we need this library to be built in hpc stack in both regular and debug mode to test the new features in this esmf version. We need the lib on hera/orion/dell

rsdunlapiv:
ESMF snapshots come out fairly frequently, so we should have a process to request this.  Should we plan to create an issue whenever we need a new snapshot built?

Hang-Lei-NOAA:
I will added these to the testing versions of hpc-stack

Hang-Lei-NOAA:
They (both debug and normal of 810bs36) are added.

edwardhartnett:
This seems to be a duplicate of #77.

kgerheiser:
Yes, I missed this.

edwardhartnett:
OK, we will close this with #77.

junwang-noaa:
Ed, as Rocky asked, I am still not clear what is the  process to request
ESMF lib that is frequently changing under hpc stack?

edwardhartnett:
@junwang-noaa and @rsdunlapiv we are about to do a 1.1.0 release of hpc-stack which supports both of the ESMF versions you require.

I will open a separate issue about how we may best support multiple, and rapidly changing, ESMF releases.

For now, yes, please open an issue every time you need a new ESMF snapshot installed.

edwardhartnett:
I believe this may be closed now.

PR #87, opened 2020-11-19
Update ESMF to bs36

### LINKED ISSUE #77: Update ESMF in develop branch?
In the develop branch there's ESMF bs27, while in the branch `feature/nco` bs36 is being used. 

Should we update?

kgerheiser:
Fixes #77 #41

### DIFF
--- a/config/stack_custom.yaml
+++ b/config/stack_custom.yaml
@@ -71,7 +71,7 @@ pio:
 
 esmf:
   build: YES
-  version: 8_1_0_beta_snapshot_27
+  version: 8_1_0_beta_snapshot_36
   shared: YES
   enable_pnetcdf: NO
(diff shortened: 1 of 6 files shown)


aerorahul:
I think you need to update the `stack_gaea.yaml` file too and possibly the weather ci.
We have way too many files with the same YES|NO for noaa platforms.
Can we see if we can combine noaa, gaea (atleast) and may be even ncar.  The weather ci one disables JEDI components, so I am not sure how to handle that.

Thoughts?

kgerheiser:
Ah yes, new stack files have been added.

Yeah, I think there has been a proliferation of these stack files that are identical.

aerorahul (approved):

Issue #111, opened 2020-11-24
Version 1.1.0 Installation and test status


kgerheiser:
This issue is to track the test and installation status of hpc-stack on NOAA systems.

**Note.** The following table applies to the second pre-release tag 1.1.0, created 12/02/2020. The comments below, up to https://github.com/NOAA-EMC/hpc-stack/issues/111#issuecomment-738048057, relate to the first pre-release tag.

Platform | Installation Status | Tested with UFS Regression tests
-------- | ------------------ |--------
Orion     |      Installed (`/apps/contrib/NCEP/libs/hpc-stack/modulefiles/stack`)     | Pass with Intel
Hera      |     Installed   (`/scratch2/NCEPDEV/nwprod/hpc-stack/libs/hpc-stack/modulefiles/stack`)      | Pass with Intel and GNU
Jet         | Installed (`lfs4/HFIP/hfv3gfs/nwprod/hpc-stack/libs/modulefiles/stack`)             | Pass with Intel
Dell        | Installed (`/usrx/local/nceplibs/dev/hpc-stack/libs/hpc-stack/modulefiles/stack`)             | Pass with Intel
Cray      | NCEPLIBS v1.3.0 installed (`/usrx/local/nceplibs/NCEPLIBS/cmake/install/NCEPLIBS-v1.3.0/modules`)           | Pass with Intel
Gaea     |  Installed (`/lustre/f2/pdata/esrl/gsd/ufs/hpc-stack-v1.1.0`)            | Pass with Intel
Cheyenne | Installed (`/glade/p/ral/jntp/GMTB/tools/hpc-stack-v1.1.0`)         | Pass with Intel and GNU
macOS |     N/A                      |  Tested on macOS Mojave with LLVM clang 9.0.0 + GNU gfortran 9.2.0 + pre-installed mpich-3.3.1, using a canned test case and a recent hash (bfce1df1c03c3aef87c177cd0b04579d0ce5f430) of the ufs-weather-model

climbfuji:
Additional systems that have been tested:

- macOS Catalina with gcc-9.2.0/gfortran-9.2.0 and pre-installed mpich-3.3.1: installed, not yet tested

[...]

PR #112, opened 2020-11-30
Install ESMF beta 21 on macOS

kgerheiser:
ESMF beta 27 does not work on macOS. Update `stack_mac` to install beta 21.

https://github.com/ufs-community/ufs-weather-model/issues/303

https://github.com/NOAA-EMC/hpc-stack/issues/111#issuecomment-735869981

### DIFF
--- a/config/stack_mac.yaml
+++ b/config/stack_mac.yaml
@@ -71,7 +71,7 @@ pio:
 
 esmf:
   build: YES
-  version: 8_1_0_beta_snapshot_27
+  version: 8_1_0_beta_snapshot_21
   shared: YES
   enable_pnetcdf: NO
   debug: NO
--- a/README.md
+++ b/README.md
@@ -377,6 +377,11 @@ setenv("LMOD_EXACT_MATCH", "no")
 setenv("LMOD_EXTENDED_DEFAULT", "yes")
 ```
 
+## Known Issues
+
+- ESMF beta snapshot 27 does not work on macOS. `stack_mac` installs
+  beta 21 instead.
+
 ## Disclaimer
 
 The United States Department of Commerce (DOC) GitHub project code is


edwardhartnett:
Is this for the 1.1.0 release?

edwardhartnett:
Sorry, dumb question, of course it is...

climbfuji:
> Sorry, dumb question, of course it is...

You beat me for the answer!

config/stack_mac.yaml
```diff
@@ -71,7 +71,7 @@ pio:
 
 esmf:
   build: YES
-  version: 8_1_0_beta_snapshot_27
+  version: 8_1_0_beta_snapshot_21
   shared: YES
```
climbfuji, line 75: Can you remind me why are we building the shared library at all?

climbfuji (commented):
Please note that there will be a follow-up PR from my side with updates to `libs/build_esmf.sh`.

config/stack_mac.yaml
kgerheiser: I don't have a good answer. Because shared libraries are the standard on macOS? JEDI related? Maybe @aerorahul has an answer.

kgerheiser:
Do you just want to put this into your PR, then? Instead of having two?

config/stack_mac.yaml
aerorahul: On a macOS. Yes. On NOAA platforms it is static. 
I have no preference. Just want to test shared somewhere. I prefer shared, in general. And seems macOS is a niche club.

climbfuji: I usually build the static library only, it's cleaner and you avoid potential `rpath` issues on macOS.

climbfuji:
> Do you just want to put this into your PR, then? Instead of having two?

We can have two, no problem. Your's first, then mine.

climbfuji (approved):

PR #115, opened 2020-11-30
Add an option to build NetCDF-C++. It fails to build with LLVM Clang

### LINKED ISSUE #114: Add an option to disable building NetCDF C++
NetCDF C++ fails to build with LLVM Clang:

https://github.com/NOAA-EMC/hpc-stack/issues/111#issuecomment-733280376

kgerheiser:
Fixes #114

### DIFF
--- a/libs/build_netcdf.sh
+++ b/libs/build_netcdf.sh
@@ -12,6 +12,7 @@ compiler=$(echo $HPC_COMPILER | sed 's/\//-/g')
 mpi=$(echo $HPC_MPI | sed 's/\//-/g')
 
 [[ ${STACK_netcdf_enable_pnetcdf:-} =~ [yYtT] ]] && enable_pnetcdf=YES || enable_pnetcdf=NO
+[[ ${STACK_netcdf_disable_cxx:-} =~ [yYtT]  ]] && enable_cxx=NO || enable_cxx=YES
 
 if $MODULES; then
     set +x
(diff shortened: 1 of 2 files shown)


climbfuji (approved):
Haven't tested it, but the changes look good!

config/stack_custom.yaml
```diff
@@ -51,6 +51,7 @@ netcdf:
   build: YES
   shared: YES
   enable_pnetcdf: NO
+  build_netcdf_cxx: YES
```
aerorahul: Instead of adding this to every yaml file, the default is YES to build netcdf_cxx
Just define `disable_netcdf_cxx: YES` and in the `build_netcdf.sh`check if that is Y.  If not, proceed as usual.

config/stack_mac.yaml
```diff
@@ -51,6 +51,7 @@ netcdf:
   build: YES
   shared: YES
   enable_pnetcdf: NO
+  build_netcdf_cxx: YES
```
aerorahul: This needs to be changed to `disable_cxx`
Also in the Known issues of the `README.md`

aerorahul: Also, since `YES` is the default, we don't really need to set it explicitly to YES.  It is only when one wishes to disable it, it needs to be told to do so.

aerorahul: In other words, remove this line.

libs/build_netcdf.sh
```diff
@@ -174,28 +175,31 @@ VERBOSE=$MAKE_VERBOSE make -j1 #NetCDF-Fortran-4.5.2 & intel/20 have a linker bu
 [[ $MAKE_CHECK =~ [yYtT] ]] && make check
 $SUDO make install
 
-cd $curr_dir
+if [ $enable_cxx =~ [yYtT] ]; then 
+   
+  cd $curr_dir
 
-$MODULES || echo $software >> ${HPC_STACK_ROOT}/hpc-stack-contents.log
+  $MODULES || echo $software >> ${HPC_STACK_ROOT}/hpc-stack-contents.log
```
aerorahul: actually, you need to move this line above the if-block.

aerorahul (approved):
looks good.
Thanks!

PR #124, opened 2020-12-08
Fix disable_cxx option

### LINKED ISSUE #123: Netcdf c++ not built even though disable_cxx config is set to NO in stack_<>.yaml file.
**Describe the bug**
The netcdf c++ package is a required dependency for JCSDA's soca-science compile.  However, if one sets the disable_cxx setting under the netcdf section to NO, the build will not compile the netcdf c++ package.

**To Reproduce**
1) Include the 'disable_cxx: NO' in the stack config yaml.
2) Compile the hpc-stack using the ./build_stack.sh command.
3) Once it is complete, verify the the libnetcdf_c++4.a and libnetcdf_c++4.la do not exist in the expected location.

**Expected behavior**
The netcdf c++ compile should be executed, resulting in funnctional libnetcdf_c++4.a and libnetcdf_c++4.la .

**System:**
Ubuntu 20.04

**Additional context**
Attempts to compile JCSDA's soca-science (which requires netcdf c++4) failed to complete the cmake configuration step.

The error during hpc-stack compile, just after the netcdf fortran compile looks like the following...
+ false
+ echo netcdf-fortran-4.5.3
+ '[' YES '=~' '[yYtT]' ']'
/hpc-stack/libs/build_netcdf.sh: line 180: [: =~: binary operator expected

And then it just bails from the script after that.

** Fix **
the fix on line 180 of build_netcdf.sh is
if [[ ${enable_cxx} =~ [yYtT] ]]; then

kgerheiser:
Need double brackets around variable. Tested building with and without. 

Fixes #123

### DIFF
--- a/libs/build_netcdf.sh
+++ b/libs/build_netcdf.sh
@@ -12,7 +12,7 @@ compiler=$(echo $HPC_COMPILER | sed 's/\//-/g')
 mpi=$(echo $HPC_MPI | sed 's/\//-/g')
 
 [[ ${STACK_netcdf_enable_pnetcdf:-} =~ [yYtT] ]] && enable_pnetcdf=YES || enable_pnetcdf=NO
-[[ ${STACK_netcdf_disable_cxx:-} =~ [yYtT]  ]] && enable_cxx=NO || enable_cxx=YES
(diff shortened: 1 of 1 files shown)


aerorahul (approved):
looks good to me.
Thanks for fixing the bug.

climbfuji (approved):

Issue #122, opened 2020-12-07
Issue in build_nceplibs.sh at nemsio compile step - mpi local variable not set for "no modules" configuration


slawrence87544:
During the setup for the build_nceplibs.sh script, the local variable 'mpi' is not set to local variable 'mpi_check' as it is in the "use modules" case.  Several ncep libraries rely on this local variable 'mpi' downstream to set the the compile flags (FC, CC, CXX).

**To Reproduce**
When calling './build_stack.sh' omit the -m flag to keep the 'MODULES' environment variable set to 'false'.  Ensure that nemsio nceplib is configured to be built in the stack_<>.yaml config file.

**Expected behavior**
The hpc-stack build will fail during the nemsio build step.

**System:**
Ubuntu 20.04

**Additional context**
Expected error...

+ git submodule update --init --recursive
+ cd ..
+ [[ N =~ [yYtT] ]]
+ [[ -d nemsio-v2.5.2 ]]
+ cd nemsio-v2.5.2
+ [[ -d build ]]
+ mkdir -p build
+ cd build
+ cmake .. -DCMAKE_INSTALL_PREFIX=/home/builder/opt -DENABLE_TESTS=OFF -DOPENMP=OFF
-- The Fortran compiler identification is Intel 20.2.1.20201112
-- Check for working Fortran compiler: /opt/intel/oneapi/compiler/2021.1.1/linux/bin/intel64/ifort
-- Check for working Fortran compiler: /opt/intel/oneapi/compiler/2021.1.1/linux/bin/intel64/ifort  -- works
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Checking whether /opt/intel/oneapi/compiler/2021.1.1/linux/bin/intel64/ifort supports Fortran 90
-- Checking whether /opt/intel/oneapi/compiler/2021.1.1/linux/bin/intel64/ifort supports Fortran 90 -- yes
-- Setting build type to 'Release' as none was specified.
CMake Error in /tmp/hpc-stack/pkg/nemsio-v2.5.2/build/CMakeFiles/CMakeTmp/CMakeLists.txt:
  Imported target "MPI::MPI_Fortran" includes non-existent path
    "/tmp/hpc-stack/pkg/nemsio-v2.5.2/'/opt/intel/oneapi/mpi/2021.1.1/include'"
  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
[...]

PR #125, opened 2020-12-08
fix: mpi local variable not set for "no modules" configuration

slawrence87544:
body: During the setup for the build_nceplibs.sh script, the
local variable 'mpi' is not set to local variable 'mpi_check'
as it is in the "use modules" case. Several ncep libraries rely
 on this local variable 'mpi' downstream to set the the compile
 flags (FC, CC, CXX).

This pull request fixes issue #122

id: #122


kgerheiser (approved):

aerorahul (approved):

kgerheiser:
Thanks for the PR and bug report

slawrence87544:
No problem, happy to help contribute.  It's my chance to give back to what
is a huge time saver (that is... the hpc-stack).

PR #128, opened 2020-12-11
Cron test script

### LINKED ISSUE #81: set up cron job on hpc systems
cron jobs can test hpc-stack and email the results.

kgerheiser:
`setup_cron.sh` sets variables and is the entry point of cron. It then downloads hpc-stack and runs `build-and-test.sh` with the options given by `setup-cron.sh`.

Then, `build-and-test.sh` builds the develop branch hpc-stack (along with the debug version of ESMF), and then it checks out the ufs-weather-model and runs the regression test.

A hash is saved after each build so that it is only re-built when there are changes. 

A nicely formatted email is sent out with the results of the test, for example:

```
==============================================================================
Mon Dec 14 03:33:19 UTC 2020
hpc-stack hash: bbbea9dec4768febe3feb7a71a0b1268a5cc80b7
ufs hash: 9bd6b59b3c5a230558430f238e32ca60d2844e57

hpc-stack build: SUCCESS
hpc-stack log: /home/Kyle.Gerheiser/.hpc-stack/logs/hpc-stack_12-14-2020-00:53.log
UFS regression test: PASS
ufs log: /home/Kyle.Gerheiser/.hpc-stack/logs/ufs_12-14-2020-02:29.log

==============================================================================
```

Fixes #81


cron-ci/README.md
```diff
@@ -0,0 +1,15 @@
+## Cron CI
+
+Provides a script that can be run using CRON on HPC systems to build and test hpc-stack.
+
+Saves a hash of the last build and does not build if the develop branch hasn't changed
+
+## Variables
+
+* HPC_HOMEIDR - Path to store logs and save temporary data
```
aerorahul: `HOMEDIR` not HOMEIDR, lol!

edwardhartnett (approved):
Awesome!

kgerheiser:
Ready for review now.

I tested it out on Hera over the weekend. I fixed some odds and ends, added the UFS tests, and added a nicely formatted email with the results. 

I'll install it across our machines this week. 

It will be installed in a public `nightly-develop` folder every day (if there are changes in hpc-stack) alongside the versioned release so that anyone can try out the latest and greatest.

edwardhartnett (approved):

cron-ci/build-and-test.sh
```diff
@@ -0,0 +1,126 @@
...
+       sed -i "s:module load \(.*\)/.*:module load \1:" modulefiles/${machine}.intel/*
+
+       # give version to ESMF because two versions are built
+       sed -i "s:module load esmf:module load esmf/${STACK_esmf_version}:" modulefiles/${machine}.intel/fv3
+       sed -i "s:module load esmf:module load esmf/${STACK_esmf_version}-debug:" modulefiles/${machine}.intel/fv3_debug
+
+       cd tests
+       ./rt.sh -f -e >> ${ufs_log} 2>&1
+   fi
+fi
```
aerorahul: I suggest we put this (TEST_UFS) in another script specifically for UFS.
In the future, we can add as many applications to build and test e.g. JEDI.

Keep this script to build the stack alone.

cron-ci/setup-cron.sh
```diff
@@ -0,0 +1,41 @@
...
+cd $HPC_DOWNLOAD_PATH
+
+# delete hpc-stack source if it exists
+rm -rf hpc-stack
+rm -rf ${HPC_INSTALL_PATH}/*
+
+git clone https://github.com/NOAA-EMC/hpc-stack.git > /dev/null 2>&1
+cd hpc-stack
+
+./cron-ci/build-and-test.sh
```
aerorahul: Add a second script to test applications.
e.g.
```
./cron-ci/build-hpc-stack.sh
rc=$?
[[ $rc -ne 0 ]] && ( echo "building of hpc-stack failed, ABORT!" )
[[ $rc -ne 0 ]] && exit 1

if [[ $TEST_UFS == true ]]; then
    ./cron-ci/build-ufs.sh
fi
```

aerorahul (changes requested):
some suggestions for modularizing

kgerheiser:
How about this? 

I factored out the testing into its own script, and created another script `test-applications.sh` which calls the individual test scripts.

cron-ci/build-hpc-stack.sh
```diff
@@ -0,0 +1,73 @@
...
+    echo ""
+
+     # check if hpc-stack build succeded 
+    if grep -qi "build_stack.sh: SUCCESS!" ${hpc_log}; then
+        echo "hpc-stack build: PASS"
+    else
+        echo "hpc-tack build: FAIL"
+        exit 1
+    fi
+fi
```
aerorahul: While this is perfectly acceptable, I prefer not using big if-else-blocks.  Here is an alternative:
```
[[ "$build" == false ]] && ( echo "Nothing changed, so nothing to build"; exit 0 )

# Don't build these libs because NOAA RDHPCS systems can't use wget or connect to gitlab
# finds the name of the library and changes the next line to "build: NO"
sed -i '/^udunits/{n;s/.*/  build: NO/}' ${HPC_STACK_FILE}
sed -i '/^szip/{n;s/.*/  build: NO/}' ${HPC_STACK_FILE}
...
```

aerorahul (changes requested):
I like this much better.
One suggestion is to also turn on `set -x`.  It produces a lot of output, but is worth it for debugging.
Do however `set +x` around any `module` commands.

cron-ci/build-hpc-stack.sh
```diff
@@ -0,0 +1,73 @@
...
+    echo ""
+    echo "hpc-stack hash: ${current_hash}"
+    echo "hpc-stack log: ${hpc_log}"
+    echo ""
+
+     # check if hpc-stack build succeded 
+    if grep -qi "build_stack.sh: SUCCESS!" ${hpc_log}; then
+        echo "hpc-stack build: PASS"
+    else
+        echo "hpc-tack build: FAIL"
```
aerorahul: typo `hpc-stack`.

kgerheiser:
I removed the if-block and fixed the typo.

I added a special return code `build-hpc-stack.sh` if the stack isn't built because the hash hasn't changed so that the top-level script knows not to run the tests. Is there a better way to handle that case?

I enabled `-x` in `build-hpc-stack.sh` and redirected it to a log. 

I don't want to get an email everyday with all that output. It's available in the logs.

cron-ci/build-hpc-stack.sh
```diff
@@ -0,0 +1,69 @@
...
+        build=false
+        echo `date`
+        echo ""
+        echo "hpc-stack version has not changed since last time. Not building."
+        echo "hpc-stack hash: ${current_hash}"
+    fi
+fi
+
+# Return 42 if not built because hash value hasn't changed
+[[ "$build" == false ]] && ( echo "Nothing has changed, so nothing to build"; exit 42)
```
aerorahul: why 42?
This is not an error.  Non-zero return codes are treated as errors.

cron-ci/build-hpc-stack.sh
```diff
@@ -0,0 +1,69 @@
...
+
+# check if hpc-stack build succeded 
+if grep -qi "build_stack.sh: SUCCESS!" $1; then
+    echo "hpc-stack build: PASS"
+else
+    echo "hpc-stack build: FAIL"
+    exit 1
+fi
+
+
```
aerorahul, line 30: `exit 0` here.

cron-ci/setup-cron.sh
```diff
@@ -0,0 +1,57 @@
...
+git clone https://github.com/NOAA-EMC/hpc-stack.git > /dev/null 2>&1
+cd hpc-stack
+
+./cron-ci/build-hpc-stack.sh ${hpc_log} >> ${hpc_log} 2>&1
+
+rc=$?
+
+# rc == 42 - hpc-stack didn't build because hash value hasn't changed
+if [[ $rc -eq 42 ]]; then
+    exit 0
```
aerorahul: don't really need this.

cron-ci/build-hpc-stack.sh
kgerheiser: Do you have a better suggestion of a way to indicate to the parent script that hpc-stack wasn't built, as to not run the tests?

Or just run the tests anyway?

aerorahul: I see what you mean.

kgerheiser: I think the simple answer is to just move that check up to the top-level script.

cron-ci/build-hpc-stack.sh
```diff
@@ -0,0 +1,69 @@
...
+if [[ -f "${HPC_HOMEDIR}/prev_hash.txt" ]]; then
+    prev_hash=$(cat ${HPC_HOMEDIR}/prev_hash.txt)
+    if [[ "$current_hash" == "$prev_hash" ]]; then
+        build=false
+        echo `date`
+        echo ""
+        echo "hpc-stack version has not changed since last time. Not building."
+        echo "hpc-stack hash: ${current_hash}"
+    fi
+fi
```
aerorahul: Can you move this section to `setup-cron.sh`

kgerheiser: Yep

aerorahul: and then, if the hash is updated, call this script and the rest after than.
If it is not, exit there and then

kgerheiser:
Done

aerorahul (approved):
looks great.
Thanks for setting this up.

cron-ci/setup-cron.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash -l
```
aerorahul, line 1: I noticed you are using `-l` so that `$MODULESHOME/init/bash` is sourced because that is the case on HPC's.
When a sudo executes this script, there will be no login shell.

My recommendation is to remove `-l` and do a `source $MODULESHOME/init/sh` wherever modules are being loaded/used/etc.

kgerheiser: I'll try it out, but I don't think `$MODULESHOME` will even be set. You'll have to `source` the absolute directory like `/apps/contrib/lmod/lmod/init/bash` which is a pain to set for each machine.

PR #129, opened 2020-12-18
Cron-CI improvements

kgerheiser:
Group logs into folders based on date

Move UFS logging into `test-ufs.sh`

Check if `fv3_debug` exists before running sed on it. Jet, for example, does not have it and it causes an error.

Change how sed is run on `fv3` modules. Only remove the versions between `module load hpc` and end-of-file as to not affect non hpc-stack module loads.

Tidy up output formatting

### DIFF
--- a/cron-ci/test-ufs.sh
+++ b/cron-ci/test-ufs.sh
@@ -1,24 +1,41 @@
-#!/bin/bash -l
-set -eux
+#!/bin/bash
+set -eu
 
 cd ${HPC_DOWNLOAD_PATH}
 rm -rf ufs-weather-model
 
-git clone https://github.com/ufs-community/ufs-weather-model.git
+# mm-dd-yyy-hh:mm
+ufs_logdate=$(date +'%m-%d-%Y-%R')
+ufs_logname=ufs_${ufs_logdate}.log
+ufs_log=${HPC_LOG_PATH}/${ufs_logname}
+
+git clone https://github.com/ufs-community/ufs-weather-model.git > >> ${ufs_log}
 cd ufs-weather-model
-git submodule update --init --recursive
+
+ufs_hash=$(git -rev-parse HEAD)
+
+echo ""
+echo "testing ufs-weather-model..."
+echo ""
+echo "UFS log: ${ufs_log}"
+echo "UFS hash: ${ufs_hash}"
+echo ""
+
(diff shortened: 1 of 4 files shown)


kgerheiser:
Someone can approve this? I'd like to install this today.

cron-ci/test-ufs.sh
```diff
@@ -1,24 +1,41 @@
...
-cd tests
-./rt.sh -f -e
 
+# check if fv3_debug exists. sed fails if it doesn't.
+if [[ -f modulefiles/${HPC_MACHINE_ID}.intel/fv3_debug ]]; then
+    sed -i "s:module load esmf:module load esmf/${STACK_esmf_version}-debug:" modulefiles/${HPC_MACHINE_ID}.intel/fv3_debug
+fi
 
+cd tests
+./rt.sh -f -e >> ${ufs_log} 2>&1
```
climbfuji, line 41: Note that for the GNU runs, you need to use `-l rt_gnu.conf` instead of `-f`. Some machines (wcoss2, stampede) have their own `rt_MACHINE.conf` files, too.

climbfuji (commented):
I am certainly not qualified to approve this PR, but besides the specific comment below I am wondering how the environment variables `ACCNR` and `RT_COMPILER` get/can be set. `ACCNR` is required if the default account in `rt.sh` for the machine doesn't match the user. `RT_COMPILER` is required to switch from (default) Intel to GNU.

cron-ci/test-ufs.sh
kgerheiser: These are helpful comments.

You can set custom environment variables like ACCNR in the top-level script `setup-cron.sh`. For example, I set `LMOD_EXACT_MATCH=NO` for WCOSS Dell. I'll add some notes.

Right now, I have it setup to only build for Intel. I'll add a variable for specifying the `rt.conf` file. Is Do you use `-l` instead of `-f`? When I've done it I've used `-f` along with `-l`.

climbfuji: Should be either `-f` or `-l myrt.conf` - if the combination of both works, then that's surprising (and probably a bug).

kgerheiser:
Going to close this for now. I'm going to add Dom's suggestion and expand it to use Intel and GCC.

PR #130, opened 2020-12-18
Add option to build with multiple compilers/MPIs at once

kgerheiser:
HPC_COMPILER has become HPC_COMPILERS and HPC_MPI has become HPC_MPIS, and are now arrays, and setup_modules and build_stack will loop through and compile the libraries for each all at once.

Add a BUILD_MPIS option so that the flavor/version of MPI to build is set through HPC_MPI variable and controlled by setting YES/NO for the corresponding entry in BUILD_MPIS

Thoughts on this @aerorahul?

Now you can build each compiler variation all at once instead of having to re-run the setup and build scripts.

I did this so I could build Intel and GNU for my Cron testing scripts, but I think it will be useful in general.

### DIFF
--- a/build_stack.sh
+++ b/build_stack.sh
@@ -80,130 +80,165 @@ else
   exit 1
 fi
 
-# ==============================================================================
-
-compilerName=$(echo $HPC_COMPILER | cut -d/ -f1)
-compilerVersion=$(echo $HPC_COMPILER | cut -d/ -f2)
-
-mpiName=$(echo $HPC_MPI | cut -d/ -f1)
-mpiVersion=$(echo $HPC_MPI | cut -d/ -f2)
-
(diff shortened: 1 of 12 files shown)


aerorahul:
I think there is a simpler way than this to loop over multiple compilers and mpi's and minimal edits.

kgerheiser:
I think it's pretty simple. I really only added a loop around setup_modules and build_stack, along with some checks.

aerorahul:
I can show you even simpler.

aerorahul:
see a mock up in: https://github.com/NOAA-EMC/hpc-stack/tree/feature/multi-compiler-mpi

kgerheiser:
That is simple :)

PR #132, opened 2020-12-22
Create loop_stack.sh to build multiple compiler/MPI combinations at once

kgerheiser:
I took @aerorahul's suggestion and modified the way I built multiple compiler/MPI combinations.

`loop_stack.sh` looks for `HPC_COMPILERS` and `HPC_MPIS` bash arrays to be set in the given config file and then loops over them.

Add these lines to the config file:
```
HPC_COMPILERS=("gnu/9.3.0" "intel/2020.1")
HPC_MPIS=("openmpi/4.0.1" "mpich/3.3.1")

# Optionally set HPC_YAMLS if you want to use a different yaml for each combination
#HPC_YAMLS=("stack_noaa.yaml" "stack_custom.yaml")
```

To build:
```
./loop_stack.sh -c config/config_custom.sh -y config/stack_custom.sh -p <prefix>
```

Using `loop_stack.sh` always uses modules. Multiple combinations cannot be built in a flat structure.

The same yaml file can be used for all `HPC_COMPILERS` by passing `-y` to `loop_stack.sh`, or `HPC_YAMLS` can be set in the config file as an array to use a different stack for each one. For example, you might want to build MPI with GNU, but not with Intel.

Add a `-O` (overwrite)  option to `setup_modules.sh` to disable asking whether to overwrite modules because calling `setup_modules.sh` twice will create two `hpc` modules and it eases automation.

### DIFF
--- a/config/config_orion.sh
+++ b/config/config_orion.sh
@@ -1,14 +1,11 @@
 #!/bin/bash
 
+export HPC_COMPILERS=("intel/2018.4" "intel/2019.5" "intel/2020" "gcc/8.3.0")
+export HPC_MPIS=("impi/2018.4" "impi/2019.6" "impi/2020" "openmpi/4.0.2")
+
 # Compiler/MPI combination
-#export HPC_COMPILER="intel/2019.5"
-#export HPC_MPI="impi/2019.6"
-#export HPC_COMPILER="intel/2018.4"
-#export HPC_MPI="impi/2018.4"
 export HPC_COMPILER=${HPC_COMPILER:-"intel/2020"}
 export HPC_MPI=${HPC_MPI:-"impi/2020"}
-#export HPC_COMPILER="gcc/8.3.0"
-#export HPC_MPI="openmpi/4.0.2"
 
 # Build options
 export USE_SUDO=N
--- a/config/config_orion.sh
+++ b/config/config_orion.sh
@@ -5,8 +5,8 @@
 #export HPC_MPI="impi/2019.6"
 #export HPC_COMPILER="intel/2018.4"
 #export HPC_MPI="impi/2018.4"
-export HPC_COMPILER="intel/2020"
-export HPC_MPI="impi/2020"
+export HPC_COMPILER=${HPC_COMPILER:-"intel/2020"}
+export HPC_MPI=${HPC_MPI:-"impi/2020"}
 #export HPC_COMPILER="gcc/8.3.0"
 #export HPC_MPI="openmpi/4.0.2"
 
--- a/loop_stack.sh
+++ b/loop_stack.sh
@@ -18,13 +18,11 @@ usage() {
   exit 1
 }
 
-# ==============================================================================
-
-[[ $# -eq 0 ]] && usage
 
 # Defaults:
 export PREFIX="$HOME/opt"
 config="${HPC_STACK_ROOT}/config/config_custom.sh"
+yaml="${HPC_STACK_ROOT}/config/stack_custom.yaml"
 
 while getopts ":p:c:y:h" opt; do
   case $opt in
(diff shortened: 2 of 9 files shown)


config/config_orion.sh
```diff
@@ -5,8 +5,8 @@
 #export HPC_MPI="impi/2019.6"
 #export HPC_COMPILER="intel/2018.4"
 #export HPC_MPI="impi/2018.4"
```
aerorahul: remove these commented out options since we can now have an array.

config/config_orion.sh
```diff
@@ -5,8 +5,8 @@
...
-export HPC_COMPILER="intel/2020"
-export HPC_MPI="impi/2020"
+export HPC_COMPILER=${HPC_COMPILER:-"intel/2020"}
+export HPC_MPI=${HPC_MPI:-"impi/2020"}
 #export HPC_COMPILER="gcc/8.3.0"
 #export HPC_MPI="openmpi/4.0.2"
```
aerorahul: same here.

loop_stack.sh
```diff
@@ -0,0 +1,88 @@
...
+  case $opt in
+    p)
+      export PREFIX=$OPTARG
+      ;;
+    c)
+      config=$OPTARG
+      ;;
+    y)
+      single_yaml=$OPTARG
+      ;;
```
aerorahul: Do you anticipate a use case when multiple yamls may be used?

aerorahul (changes requested):
still thinking about the changes in `setup_modules.sh`

loop_stack.sh
kgerheiser: Building MPI with GCC and not with Intel

kgerheiser: It does seem a little misplaced.

In my previous iteration of this idea I had an option to build MPI based on the version specified in the config rather than using the stack file. It seems repetitive to specify building MPI in the stack file and also specifying it in the config file. All the information regarding MPI flavor/version is available in the config. You just need to know whether to build it or not.

It also leads to the possibility of mismatching them by specifying it in two places. You could change what MPI version you're building in the stack file, but not in the config (which I've done before).

My idea was to have another array `BUILD_MPIS=(NO NO YES)`, for example. Not really a fan of that either, but it works. It needs to come from somewhere.

kgerheiser: I removed it. After thinking about it, I don't think it's necessary.

kgerheiser:
The only question that remains is the overwrite option in `setup_modules.sh`

loop_stack.sh
```diff
@@ -0,0 +1,77 @@
...
+  echo "  -c  use configuration file <config> DEFAULT: config/config_custom.sh"
+  echo "  -y  use yaml file <yaml>            DEFAULT: config/stack_custom.yaml"
+  echo "  -h  display this message and quit"
+  echo
+  exit 1
+}
+
+# ==============================================================================
+
+[[ $# -eq 0 ]] && usage
```
aerorahul: That is not sufficient.  You also need a yaml file, since you don't define a default.  It is in the usage, but not in the script.

loop_stack.sh
```diff
@@ -0,0 +1,77 @@
...
+
+# Defaults:
+export PREFIX="$HOME/opt"
+config="${HPC_STACK_ROOT}/config/config_custom.sh"
+
```
aerorahul, line 26:
```
yaml="${HPC_STACK_ROOT}/config/stack_custom.yaml"
```

loop_stack.sh
aerorahul: Actually, you don't need this check at ll, since all the default options are pre-defined.

PR #134, opened 2020-12-22
Get Udunits from Github and build with CMake

### LINKED ISSUE #133: UDUNITS 2.2.26 no longer available
Auto build failed:
https://github.com/NOAA-EMC/hpc-stack/pull/132/checks?check_run_id=1595924507

Checking the Unidata website:
https://artifacts.unidata.ucar.edu/service/rest/repository/browse/downloads-udunits/

2.2.28 is the only version available at that location

kgerheiser:
Fixes #133

### DIFF
--- a/libs/build_udunits.sh
+++ b/libs/build_udunits.sh
@@ -37,19 +37,23 @@ export FCFLAGS="$FFLAGS"
 cd ${HPC_STACK_ROOT}/${PKGDIR:-"pkg"}
 
 software=$name-$version
-url=ftp://ftp.unidata.ucar.edu/pub/udunits/$software.tar.gz
-[[ -d $software ]] || ( $WGET $url; tar xvf $software.tar.gz && rm -f $software.tar.gz )
+gitTag="v${version}"
+gitURL=https://github.com/Unidata/UDUNITS-2.git
+
+[[ -d $software ]] || ( git clone -b $gitTag $gitURL $software)
 [[ ${DOWNLOAD_ONLY} =~ [yYtT] ]] && exit 0
 [[ -d $software ]] && cd $software || ( echo "$software does not exist, ABORT!"; exit 1 )
 [[ -d build ]] && rm -rf build
 mkdir -p build && cd build
 
-[[ $enable_shared =~ [yYtT] ]] && shared_flags="" || shared_flags="--disable-shared --enable-static"
+[[ $enable_shared =~ [yYtT] ]] && shared_flags=ON || shared_flags=OFF
 
-../configure --prefix=$prefix $shared_flags
+cmake .. \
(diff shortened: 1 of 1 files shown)


libs/build_udunits.sh
```diff
@@ -37,19 +37,22 @@ export FCFLAGS="$FFLAGS"
...
-[[ $enable_shared =~ [yYtT] ]] && shared_flags="" || shared_flags="--disable-shared --enable-static"
+[[ $enable_shared =~ [yYtT] ]] && shared_flags=ON || shared_flags=OFF
 
-../configure --prefix=$prefix $shared_flags
+cmake .. \
+      -DCMAKE_INSTALL_PREFIX=$prefix \
+      -DBUILD_SHARED_LIBS=$shared_flags
 
 make -j${NTHREADS:-4}
-[[ "$MAKE_CHECK" = "YES" ]] && make check
```
aerorahul, line 52: If there is a `ctest` option, we should run that in place of `make check`.  But, a check option should exist.

aerorahul (changes requested):
changes are good, but I have beef against removing the `make check` option. See comment.

libs/build_udunits.sh
kgerheiser: Re-added, but `make test` is the correct option.

It won't run any tests though unless you have some C unit testing framework.

libs/build_udunits.sh
```diff
@@ -37,19 +37,23 @@ export FCFLAGS="$FFLAGS"
...
+[[ "$MAKE_CHECK" = "YES" ]] && make test
```
aerorahul, line 56: Does `make test` run the tests or just builds the executables for the tests?

kgerheiser: It runs the tests, as is usual with CMake.

```
> make test
Running tests...
Test project /Users/KIG/Desktop/blarn/hpc-stack/pkg/udunits-2.2.26/build
No tests were found!!!
```

libs/build_udunits.sh
aerorahul: how to do that?

kgerheiser: You need CUnit. When you run CMake it looks for it:

`-- Unable to find CUnit library. Unit-testing not enabled.`

kgerheiser: The reason I went with CMake was because the source code does not have `configure` file to run. You have to generate it yourself using `autoconf` and if you don't have all those Autotools installed you won't get a `configure` script to run.

aerorahul: sounds good. thanks for the info.

aerorahul (approved):

PR #143, opened 2021-01-21
bugfix/pio tag names

### LINKED ISSUE #144: PIO has an inconsistency in its release tag. Adapt to it.
**Describe the bug**
PIO had an inconsistency with the naming of tag `2_5_1`.
https://github.com/NCAR/ParallelIO/issues/1711
They are now back to regular with `2_5_2`.

**To Reproduce**
use `develop`

**Expected behavior**
PIO will fail.

aerorahul:
PIO repo has an aberration in the release of tag `2_5_1`.  The tag is named `pio_2_5_1` as opposed to the convention for all other tags which is `pioX_Y_Z`.

This bugfix is only for `2_5_1`.  For all other tags, the version should be `X.Y.Z`

closes #144

### DIFF
--- a/config/stack_custom.yaml
+++ b/config/stack_custom.yaml
@@ -69,7 +69,7 @@ cdo:
 
 pio:
   build: YES
-  version: _2.5.1
+  version: 2.5.1
   enable_pnetcdf: NO
   enable_gptl: NO
 
--- a/config/stack_custom.yaml
+++ b/config/stack_custom.yaml
@@ -69,7 +69,7 @@ cdo:
 
 pio:
   build: YES
-  version: 2.5.1
+  version: _2.5.1
   enable_pnetcdf: NO
   enable_gptl: NO
 
--- a/config/stack_gaea.yaml
+++ b/config/stack_gaea.yaml
@@ -66,7 +66,7 @@ cdo:
 
 pio:
   build: YES
-  version: 2.5.1
+  version: _2.5.1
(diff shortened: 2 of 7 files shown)


kgerheiser (approved):

aerorahul:
@kgerheiser 
I think we can do better.

kgerheiser:
In what way

aerorahul:
@kgerheiser 
we do a bunch of `sed` stuff to map `X.Y.Z` into `X_Y_Z` that PIO names.
We could just use `X_Y_Z`
That still does not address the aberration in the PIO naming convention for `2_5_1`.
This bugfix will create a modulefile `_2.5.1`, which is not very nice.

So, I think the best way is to address it with an `if-block` for `2_5_1`.  Its unfortunate, unless PIO correct their error and re-release `2_5_1` correctly.

Thoughts?

Hang-Lei-NOAA:
Yes, That is good.
If they change, we can further modify.

kgerheiser:
I didn't think of the module. Yes, I think an if is the way to go. It has been done before.

edwardhartnett:
Well I hate to admit it but the PIO release team is me. ;-)

However, I hate to redo an old release when I have a new one almost ready to go. I expect we will release 2.5.4 in the next few weeks. I will try and get the tag name correct!

kgerheiser (approved):

PR #145, opened 2021-01-21
feat: add support for a unified hpc-stack and jedi-stack

slawrence87544:
body: This modification of hpc-stack enables support for compiles
of both noaa-emc/gsi and jcsda-internal/soca-science as well as
the ufs weather model.  This branch also adds a stack config which
includes compile of cmake as well as it seems like compile of
the jedi repos goes more smoothly when cmake is compiled vs when
it is untarred from a pre-built package.

### DIFF
--- a/config/stack_jedi.yaml
+++ b/config/stack_jedi.yaml
@@ -61,6 +61,10 @@ nco:
   build: NO
   version: 4.9.3
 
+cdo:
+  build: YES
+  version: 1.9.8
+
 pio:
   build: YES
   version: 2.5.1
@@ -75,7 +79,7 @@ esmf:
   debug: NO
 
 fms:
-  build: YES
+  build: NO
   version: master
   repo: noaa-gfdl
   cmake_opts: "-DGFS_PHYS=ON -DLARGEFILE=ON"
@@ -215,7 +219,7 @@ eigen:
   version: 3.3.7
 
 gsl_lite:
-  build: YES
+i  build: YES
   version: 0.37.0
 
 gptl:
@@ -261,3 +265,4 @@ atlas:
   build: YES
   version: release-stable
   repo: jcsda
+
--- a/config/stack_jedi.yaml
+++ b/config/stack_jedi.yaml
@@ -209,6 +209,11 @@ grib_util:
   install_as: 1.2.2
   openmp: ON
 
+ncio:
+  build: YES
+  version: develop
+  install_as: 1.0.0
+
 boost:
   build: YES
   version: 1.68.0
--- a/config/stack_jedi.yaml
+++ b/config/stack_jedi.yaml
@@ -30,7 +30,7 @@ jasper:
 udunits:
   build: YES
   shared: NO
-  version: 2.2.28
+  version: 2.2.26
 
 hdf5:
   build: YES
--- a/config/stack_jedi.yaml
+++ b/config/stack_jedi.yaml
@@ -1,0 +1,263 @@
+cmake:
+  build: YES
+  version: 3.17.2
+
+jpeg:
+  build: YES
+  shared: NO
+  version: 9.1.0
+
+zlib:
+  build: YES
+  shared: NO
+  version: 1.2.11
+
+png:
+  build: YES
+  shared: NO
+  version: 1.6.35
+
+szip:
+  build: YES
+  shared: NO
+  version: 2.1.1
+
+jasper:
+  build: YES
+  shared: NO
+  version: 2.0.22
+
+udunits:
+  build: YES
+  shared: NO
+  version: 2.2.28
+
+hdf5:
+  build: YES
+  version: 1.10.6
+  shared: YES
+  enable_szip: NO
+  enable_zlib: YES
+
+pnetcdf:
+  build: NO
+  version: 1.12.1
+  shared: NO
+
+netcdf:
+  build: YES
+  shared: YES
+  enable_pnetcdf: NO
+  disable_cxx: NO
+  version_c: 4.7.4
+  version_f: 4.5.3
+  version_cxx: 4.3.1
+
+nccmp:
+  build: YES
+  version: 1.8.7.0
+
+nco:
+  build: NO
+  version: 4.9.3
+
+pio:
+  build: YES
+  version: 2.5.1
+  enable_pnetcdf: NO
+  enable_gptl: NO
+
+esmf:
+  build: YES
+  version: 8_1_0_beta_snapshot_27
+  shared: NO
+  enable_pnetcdf: NO
+  debug: NO
+
+fms:
+  build: YES
+  version: master
+  repo: noaa-gfdl
+  cmake_opts: "-DGFS_PHYS=ON -DLARGEFILE=ON"
+
+bacio:
+  build: YES
+  version: v2.4.1
+  install_as: 2.4.1
+
+sigio:
+  build: YES
+  version: v2.3.2
+  install_as: 2.3.2
+
+sfcio:
+  build: YES
+  version: v1.4.1
+  install_as: 1.4.1
+
+gfsio:
+  build: YES
+  version: v1.4.1
+  install_as: 1.4.1
+
+w3nco:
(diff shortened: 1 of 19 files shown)


edwardhartnett:
Can you please start an issue for this work in which you describe what you are doing?

slawrence87544:
I'm putting one in now, thanks for the reminder.
-Steve

slawrence87544:
Please refer to issue #146 which is the reason for this pull request.

kgerheiser:
Looks good to me.

kgerheiser:
@aerorahul any input on this?

I tried it out and built it just fine.

libs/build_netcdf.sh
```diff
@@ -63,7 +63,7 @@ LDFLAGS1="-L$HDF5_ROOT/lib"
 LDFLAGS2=$(cat $HDF5_ROOT/lib/libhdf5.settings | grep AM_LDFLAGS | cut -d: -f2)
 [[ $enable_pnetcdf =~ [yYtT] ]] && LDFLAGS4="-L$PNETCDF_ROOT/lib"
 if [[ ${STACK_netcdf_shared:-} != [yYtT] ]]; then
-  LDFLAGS1+=" -lhdf5_hl -lhdf5"
+  LDFLAGS1+=" -lhdf5 -lhdf5_hl -lz"
```
kgerheiser: Was there a reason to swap these?

I think it should be `-lhdf5_hl -lhdf5`. The more specific library comes first.

aerorahul: why `lz`? Isn't that captured in L67 with the `Extra libraries`?
Also, what if hdf5 was built without zlib?

aerorahul: I have a feeling that some compilers read linkers left to right and some other way.
Is that true?

libs/build_netcdf.sh
```diff
@@ -150,10 +152,10 @@ set -x
 
 if [[ ${STACK_netcdf_shared} =~ [yYtT] ]]; then
   export LIBS=$($prefix/bin/nc-config --libs)
-  export LDFLAGS+=" -L$prefix/lib"
+  export LDFLAGS+=" -L$prefix/lib -lnetcdf -lhdf5_hl -lhdf5 -lz"
```
aerorahul: These should not be required because `nc-config --libs` should be giving everything except the path to the newly build netcdf-c library

libs/build_netcdf.sh
```diff
@@ -150,10 +152,10 @@ set -x
...
 else
   export LIBS=$($prefix/bin/nc-config --libs --static)
-  export LDFLAGS+=" -L$prefix/lib -lnetcdf"
+  export LDFLAGS+=" -L$prefix/lib -lnetcdf -lhdf5 -lhdf5_hl -lz"
```
aerorahul: Should not be necessary.

libs/build_netcdf.sh
```diff
@@ -119,6 +119,8 @@ mkdir -p build && cd build
              --disable-dap \
              --enable-netcdf-4 \
              --disable-doxygen \
+             --enable-parallel4 \
+             --disable-large-file-tests \
```
aerorahul: why disable large file tests? I would think we need them if we turn `MAKE_CHECK=Y` in the `config.sh`

libs/build_netcdf.sh
```diff
@@ -110,7 +110,7 @@ software=$name-"c"-$version
 [[ -d build ]] && rm -rf build
 mkdir -p build && cd build
 
-[[ ${STACK_netcdf_shared} =~ [yYtT] ]] || shared_flags="--disable-shared"
+[[ ${STACK_netcdf_shared} =~ [yYtT] ]] && shared_flags="--enable-shared --enable-static" || shared_flags="--disable-shared --enable-static"
```
aerorahul: so are we building shared and static at all times?
If so, why not add that to the `configure` instead of duplicating `--enable-static` here for both if and else.

libs/build_hdf5.sh
```diff
@@ -60,9 +60,9 @@ software=$name-$(echo $version | sed 's/\./_/g')
 [[ -d build ]] && rm -rf build
 mkdir -p build && cd build
 
-[[ -z $mpi ]] || extra_conf="--enable-parallel --enable-unsupported"
+[[ -z $mpi ]] || extra_conf="--enable-parallel --enable-unsupported --enable-build-mode=production"
 
-[[ $enable_shared =~ [yYtT] ]] || shared_flags="--disable-shared --enable-static --enable-static-exec"
+[[ $enable_shared =~ [yYtT] ]] && shared_flags="--enable-shared --enable-static" || shared_flags="--disable-shared --enable-static --enable-static-exec"
```
aerorahul: if we want to build static at all times, `--enable-static` should move to the `configure` line.

libs/build_hdf5.sh
```diff
@@ -60,9 +60,9 @@ software=$name-$(echo $version | sed 's/\./_/g')
 [[ -d build ]] && rm -rf build
 mkdir -p build && cd build
 
-[[ -z $mpi ]] || extra_conf="--enable-parallel --enable-unsupported"
+[[ -z $mpi ]] || extra_conf="--enable-parallel --enable-unsupported --enable-build-mode=production"
```
aerorahul: I don't know what `--enable-build-mode=production` is, but that is fine to add.

config/stack_jedi.yaml
```diff
@@ -0,0 +1,263 @@
+cmake:
```
aerorahul, line 1: how is `stack_jedi` different that `stack_noaa`?
I understand a few (all) are `build: YES`, but we need that on NOAA too (all except wcoss).
So, rather than adding a `stack_jedi`, can we work with `stack_noaa`?
Thoughts?

aerorahul (changes requested):
See comments.

config/stack_jedi.yaml
slawrence87544: The main difference here from stack_noaa.yaml is building hdf5 and netcdf as shared libraries.  However, after re-checking the CMakeLists.txt file for the noaa-emc/gsi build, I'll try to add the directives FIND_HDF5_HL and FIND_HDF5 to see if that helps.  From memory, those directives were getting pre-empted by the cmake/Modules/platforms/Generic.cmake script in the master branch of the noaa-emc/gsi repo.

libs/build_netcdf.sh
slawrence87544: This is now reverted to the develop branch of noaa-emc/hpc-stack.

slawrence87544: This is now reverted to the develop branch of noaa-emc/hpc-stack

slawrence87544: That could be, yes, however I was never able to independently verify that.

slawrence87544: At any rate, I have reverted both the hdf5 and netcdf library builds to what is found in the develop branch of noaa-emc/hpc-stack.

libs/build_netcdf.sh
slawrence87544: see above comments.

libs/build_netcdf.sh
slawrence87544: see above comments.

libs/build_hdf5.sh
slawrence87544: see above.

libs/build_hdf5.sh
slawrence87544: see above.

libs/build_netcdf.sh
slawrence87544: see above comments, but all netcdf and hdf5 build script changes are reverted to noaa-emc/hpc-stack develop branch.

libs/build_netcdf.sh
slawrence87544: See above comments.

aerorahul (approved):
looks good.

PR #147, opened 2021-01-25
Cron CI improvements

kgerheiser:
Add options to UFS regression tests. Their tests have been updated and the `-f` option was removed.

Use `RT_COMPILER` (intel/gnu) variable to select which UFS modules are edited to allow for building with GNU.

Remove `git -C` because it doesn't work on all versions of git.

Remove `-l` from scripts.

Group logs by day.

I've had this testing on Hera over the weekend.

### DIFF
--- a/cron-ci/setup-cron.sh
+++ b/cron-ci/setup-cron.sh
@@ -62,21 +62,21 @@ if [[ -f "${HPC_HOMEDIR}/prev_hash.txt" ]]; then
 fi
 
 echo `date`
+echo "Machine ID: ${HPC_MACHINE_ID}"
 echo ""
 echo "building hpc-stack..."
+echo ""
+echo "hpc-stack hash: ${current_hash}"
+echo "hpc-stack log: ${hpc_log}"
+echo ""
+
 ./cron-ci/build-hpc-stack.sh >> ${hpc_log} 2>&1
 
 # check if hpc-stack build succeded 
 if grep -qi "build_stack.sh: SUCCESS!" ${hpc_log}; then
     echo "hpc-stack build: PASS"
-    echo "hpc-stack hash: ${current_hash}"
-    echo "hpc-stack log: ${hpc_log}"
-    echo ""
 else
     echo "hpc-stack build: FAIL"
-    echo "hpc-stack hash: ${current_hash}"
-    echo "hpc-stack log: ${hpc_log}"
-    echo ""
     exit 1
 fi
 
--- a/cron-ci/setup-cron.sh
+++ b/cron-ci/setup-cron.sh
@@ -20,7 +20,7 @@ export HPC_STACK_FILE=config/stack_noaa.yaml
 # set machine name (hera, jet, orion, etc) so script edits correct ufs modulefiles
 export HPC_MACHINE_ID=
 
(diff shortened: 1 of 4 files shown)


cron-ci/setup-cron.sh
```diff
@@ -19,23 +19,27 @@ export HPC_STACK_FILE=config/stack_noaa.yaml
 
 # set machine name (hera, jet, orion, etc) so script edits correct ufs modulefiles
 export HPC_MACHINE_ID=
-export LOG_PATH=${HPC_HOMEDIR}/logs
+
+today=$(date +'%m-%d-%Y')
```
aerorahul: can you make this`%Y-%m-%d`?
Will make sorting by year, month, day easier to look at.

cron-ci/setup-cron.sh
```diff
@@ -19,23 +19,27 @@ export HPC_STACK_FILE=config/stack_noaa.yaml
...
 mkdir -p ${HPC_DOWNLOAD_PATH}
 mkdir -p ${HPC_HOMEDIR}
-mkdir -p ${LOG_PATH}
+mkdir -p ${HPC_LOG_PATH}
 
 cd $HPC_DOWNLOAD_PATH
 rm -rf hpc-stack
 
 # mm-dd-yyy-hh:mm
 hpc_logdate=$(date +'%m-%d-%Y-%R')
```
aerorahul: This should also be year-month-day.

cron-ci/test-applications.sh
```diff
@@ -1,28 +1,39 @@
...
 
 cron_dir=${HPC_DOWNLOAD_PATH}/hpc-stack/cron-ci
 
 if [[ "$TEST_UFS" == true ]]; then
+
+    cd ${HPC_DOWNLOAD_PATH}
+    rm -rf ufs-weather-model
+
     # mm-dd-yyy-hh:mm
     ufs_logdate=$(date +'%m-%d-%Y-%R')
```
aerorahul: here as well.

aerorahul (commented):
Scripts look good.
I have one comment about the date string.
I think it should be `year-month-day`.  It makes sorting easier and follows how we usually do this sort of thing.

cron-ci/setup-cron.sh
kgerheiser: Ok

cron-ci/setup-cron.sh
```diff
@@ -37,7 +37,7 @@ cd $HPC_DOWNLOAD_PATH
 rm -rf hpc-stack
 
 # mm-dd-yyy-hh:mm
```
aerorahul: comment is incorrect.

cron-ci/test-applications.sh
```diff
@@ -9,7 +9,7 @@ if [[ "$TEST_UFS" == true ]]; then
     rm -rf ufs-weather-model
 
     # mm-dd-yyy-hh:mm
```
aerorahul: same comment here.

aerorahul (approved):
approving. please look at the comments too.

aerorahul (approved):

PR #154, opened 2021-02-08
Differentiate between lib and lib64 in modules

kgerheiser:
Instead of hardcoding `lib`, make it a templated value and at module installation detect `CMAKE_INSTALL_LIBDIR` and fill it into the module using `configure_file`.


stack_helpers.sh
```diff
@@ -29,8 +29,21 @@ function update_modules {
 
   cd $to_dir
   $SUDO mkdir -p $name; cd $name
-  $SUDO cp $tmpl_file $version.lua
 
+  # CMAKE_INSTALL_LIBDIR is used by some projects (i.e. lib, lib64)
+  # Detect this when installing the module so the module variables point to the correct path
```
aerorahul: can we create that `CMakeLists.txt` file here instead of adding a new file to the repo.  Its purpose is really internal here.

aerorahul:
I like the changes. 
But, I have a question.
I see you are relying on `cmake` to obtain `CMAKE_INSTALL_LIBDIR` and install the modulefile.
What do you do when cmake itself is being built and deployed as a module?

kgerheiser:
@aerorahul I added a check if building `cmake`, `gnu`, or `mpi` (the first three things built) and for those just copy the module over, otherwise CMake is available.

stack_helpers.sh
```diff
@@ -29,8 +29,34 @@ function update_modules {
...
+
+        cmake ${HPC_STACK_ROOT}/${PKGDIR:-"pkg"}/libdir_test
+    fi
+
+    # The test script outputs "cmake_install_libdir.txt" which contains the lib dir value
+    CMAKE_INSTALL_LIBDIR=$(cat ${HPC_STACK_ROOT}/${PKGDIR:-"pkg"}/libdir_test/cmake_install_libdir.txt)
+
+    # Install the module with configure_file, replacing ${CMAKE_INSTALL_LIBDIR} (and potentially other variables)
+    # with the actual value for that system
+    $SUDO cmake -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR} \
```
aerorahul, line 54: This is good.
I would just note that we will need more template variables e.g. `PYTHON_VERSION` in `bufr.lua`.
We can address that in a later PR.

aerorahul (approved):
looks good.
You will need to bring you branch up to date before CI and merge.

PR #184, opened 2021-03-04
issue template update for add new package

edwardhartnett:
I'm not sure how this will look or whether this is what we really will want, because the checklist is awful long.

Anyway, let's get this merged and we can try it with a fake request, and see what it's like.

### DIFF
--- a/.github/ISSUE_TEMPLATE/add_package_request.md
+++ b/.github/ISSUE_TEMPLATE/add_package_request.md
@@ -18,3 +18,80 @@ List any other related pacakges already in hpc-stack, as far as you know.
 
 **Additional context**
 Add any other context or screenshots about the package or library here.
+
+**Checklist**
+
+
+WCOSS Software request checklist
+
+
+General questions:
+
+Supervisor or sponsor of the requester
+
+Software name and version, specific URL to the software
+
+Software type - New or Upgrade existing?
+
+Justification ( List NPS models using the software)
+
+Completion Time requested
+
+Software License including all Dependency Software Licenses 
+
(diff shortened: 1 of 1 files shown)


aerorahul:
I agree the checklist is really long.  But that does not bother me.

What does bother me, is that hpc-stack should not be limited with what WCOSS approves or disapproves.
hpc-stack does not contain the package source.  It contains information on how to get the package and build it.

If WCOSS does not approve, we can create a YAML file `stack_wcoss.yaml` and not include those packages or set `build: NO` until they are approved.

edwardhartnett:
PK @aerorahul excellent point.

So shall we make clear that the checklist at the bottom is only used for WCOSS machines?

.github/ISSUE_TEMPLATE/add_package_request.md
```diff
@@ -19,10 +19,9 @@ List any other related pacakges already in hpc-stack, as far as you know.
 **Additional context**
 Add any other context or screenshots about the package or library here.
 
-**Checklist**
+**Will This Package be Needed in WCOSS Systems?**
```
aerorahul: I think the appropriate question to ask is if this package will be a part of an operational application.  If yes, then please fill out as much information below.
If not, this package will not be made available on operational platforms e.g. WCOSS.

edwardhartnett: OK...

aerorahul: just brainstorming.

aerorahul (approved):
Looks good.

PR #198, opened 2021-03-26
Build FMS as part of the stack.

aerorahul:
This PR:
- builds FMS as part of the stack
- Builds the tag 2020.04.03 used in the UFS weather model
- Provides both single (R4) and double (R8) precision libraries in the same installation

### DIFF
--- a/config/stack_noaa.yaml
+++ b/config/stack_noaa.yaml
@@ -79,7 +79,7 @@ esmf:
 
 fms:
   build: YES
-  version: 2020.04
+  version: 2020.04.03
   repo: noaa-gfdl
   cmake_opts: "-DGFS_PHYS=ON -D64BIT=ON -DOPENMP=ON"
 
--- a/config/stack_noaa.yaml
+++ b/config/stack_noaa.yaml
@@ -78,10 +78,10 @@ esmf:
   debug: NO
 
 fms:
-  build: NO
-  version: master
+  build: YES
+  version: 2020.04
   repo: noaa-gfdl
-  cmake_opts: "-DGFS_PHYS=ON -DLARGEFILE=ON"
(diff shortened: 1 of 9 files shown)


config/stack_noaa.yaml
```diff
@@ -78,10 +78,10 @@ esmf:
   debug: NO
 
 fms:
-  build: NO
-  version: master
+  build: YES
+  version: 2020.04.03
   repo: noaa-gfdl
-  cmake_opts: "-DGFS_PHYS=ON -DLARGEFILE=ON"
+  cmake_opts: "-DGFS_PHYS=ON -D64BIT=ON -DOPENMP=ON"
```
junwang-noaa, line 84: I assume this is for 64bit build, we will also have a 32bit real build?

aerorahul: `32BIT` is on by default.
We provide both single and double precision libraries in the same installation.
The UFS can link to either based on the choice of the application.
There will be minimal changes in the UFS CMakeLists.txt, namely,  the `if(FMS)` block will look like:
```
if(FMS)
  find_package(FMS REQUIRED COMPONENTS R4 R8)
  if(32BIT)
    add_library(fms ALIAS fms_r4)
  else()
    add_library(fms ALIAS fms_r8)
  endif()
endif(FMS)
```

climbfuji (approved):
Looks good to me. I am wondering though if 64bit and 32bit should both be on by default, i.e. no need to specify an cmake option in order to get both?

aerorahul:
> Looks good to me. I am wondering though if 64bit and 32bit should both be on by default, i.e. no need to specify an cmake option in order to get both?

That will require a change at NOAA-GFDL/FMS.

PR #202, opened 2021-04-01
Add config for Cheyenne with GNU-10.1.0, fix ESMF build with GNU 10

climbfuji:
This PR
- adds a new config for Cheyenne with GNU 10 (for the moment we need to have both, GNU 9 and 10)
- updates the `stack_ncar.yaml` to match the NOAA stack file more closely (but not entirely due to issues with the old nceppost and cdo with GNU 10)
- fixes a "bug" in `libs/build_esmf.sh` so that the C++/Fortran flags are passed correctly to the ESMF build

With these changes, hpc-stack builds successfully on Cheyenne with gnu-10.1.0.


aerorahul (approved):
Looks good.
Thanks for the ESMF build fix.

kgerheiser (approved):

Issue #201, opened 2021-03-31
Install develop on Cheyenne with GNU 10 - ESMF errror


climbfuji:
**Describe the bug**
This is the second of several issues I am going to create wrt installing hpc-stack develop on Cheyenne.

This time I believe it is specific to the compiler choice GNU 10.1.0 (or any gfortran >=10). Installing ESMF fails with:
```
......
  515 |             call mpi_recv  ( hs, 1, MPI_INTEGER, root, mtag, comm, &
      |                             2
Error: Type mismatch between actual argument at (1) and actual argument at (2) (INTEGER(8)/INTEGER(4)).
/glade/work/heinzell/fv3/hpc-stack-cheyenne-gnu-10.1.0/src/hpc-stack-cheyenne-10/pkg/ESMF_8_1_0_beta_snapshot_27/src/Infrastructure/IO/PIO/pio_support.F90:222:27:

  222 |              call MPI_SEND(dof,int(sdof),PIO_OFFSET_KIND,masterproc,n,comm,ierr)
      |                           1
......
  496 |                   call mpi_send ( hs, 1, MPI_INTEGER, p, mtag, comm, ier )
      |                                  2
Error: Type mismatch between actual argument at (1) and actual argument at (2) (INTEGER(8)/INTEGER(4)).
/glade/work/heinzell/fv3/hpc-stack-cheyenne-gnu-10.1.0/src/hpc-stack-cheyenne-10/pkg/ESMF_8_1_0_beta_snapshot_27/build/common.mk:3863: recipe for target '/glade/work/heinzell/fv3/hpc-stack-cheyenne-gnu-10.1.0/src/hpc-stack-cheyenne-10/pkg/ESMF_8_1_0_beta_snapshot_27/obj/objO/Linux.gfortran.64.mpt.default/pio_support.o' failed
```
I think what is needed are the gfortran-20 compatibility flags that have already been added to the NCEPLIBS, etc.

**To Reproduce**
See https://github.com/NOAA-EMC/hpc-stack/issues/200, but change `config/config_cheyenne_gnu.sh` so that the compiler module is `gnu/10.1.0`.

**Expected behavior**
No errors during installation of cdo.

**System:**
Cheyenne with GNU-10.1.0

climbfuji:
I previously said this is a pio error, but it is not - it's ESMF.

aerorahul:
[...]

Issue #149, opened 2021-01-27
GSI Reproducibility issues using hpc-stack NetCDF4 and HDF5 on WCOSS Dell machines


MichaelLueken:
**Describe the bug**
While running the GSI on Mars and Venus, the results from an executable built using hpc-stack modules don't reproduce an executable using NCO's modules.  However, if the hpc-stack built executable uses NCO's NetCDF4 and HDF5 modules, then the results reproduce those from the NCO only executable.

**To Reproduce**
1)Clone [NOAA-EMC:master](https://github.com/NOAA-EMC/GSI) as the control and then clone [MichaelLueken-NOAA:master](https://github.com/MichaelLueken-NOAA/GSI) for the experiment.
2)Under the [MichaelLueken-NOAA:master](https://github.com/MichaelLueken-NOAA/GSI) clone, update modulefiles/modulefile.ProdGSI.wcoss_d to use hpc-stack:

module load lsf/10.1
module use /usrx/local/nceplibs/dev/hpc-stack/libs/hpc-stack/modulefiles/stack
module load hpc/1.1.0
module load hpc-ips/18.0.1.163
module load hpc-impi/18.0.1
module load prod_util/1.2.2
module load bufr/11.4.0
module load ip/3.3.3
module load nemsio/2.5.2
module load sfcio/1.4.1
module load sigio/2.3.2
module load sp/2.3.3
module load w3nco/2.4.1
module load w3emc/2.7.3
module load bacio/2.4.1
module load crtm/2.3.0
module load cmake/3.10.0
module load python/3.6.3
module load netcdf/4.7.4
module load hdf5/1.10.6

The results from this test won't reproduce [NOAA-EMC:master](https://github.com/NOAA-EMC/GSI).

If you use:

module load lsf/10.1
module use /usrx/local/nceplibs/dev/hpc-stack/libs/hpc-stack/modulefiles/stack
module load hpc/1.1.0
module load hpc-ips/18.0.1.163
module load hpc-impi/18.0.1
module load prod_util/1.2.2
module load bufr/11.4.0
module load ip/3.3.3
module load nemsio/2.5.2
module load sfcio/1.4.1
module load sigio/2.3.2
module load sp/2.3.3
module load w3nco/2.4.1
module load w3emc/2.7.3
module load bacio/2.4.1
module load crtm/2.3.0
module load cmake/3.10.0
module load python/3.6.3
module load NetCDF-parallel/4.7.4
[...]

PR #204, opened 2021-04-02
Fix CDO build by separating LDFLAGS and LIBS env variables

kgerheiser:
This fixes a build problem with CDO on WCOSS Dell (and probably other machines), and potentially #201.

I think LDFLAGS is used in this manner in other libraries, and is potentially related to https://github.com/NOAA-EMC/hpc-stack/issues/149 because `-lm` would then come at the end of the link line and use the Intel math library. 

Which throws this warning
`icc: warning #10315: specifying -lm before files may supersede the Intel(R) math library and affect performance`

Though, for now I'm just leaving it at CDO.


aerorahul (commented):
This is great @kgerheiser 
I think we may need to do this elsewhere where we do this black magic;  e.g. nccmp, nco, etc.

climbfuji:
> This is great @kgerheiser
> I think we may need to do this elsewhere where we do this black magic; e.g. nccmp, nco, etc.

Testing this now on Cheyenne ...

climbfuji:
@kgerheiser @aerorahul I have a small PR for Kyle's branch: https://github.com/kgerheiser/hpc-stack/pull/1/files

This removed stack_ncar (now identical to stack_noaa, because Kyle's bugfix works) and updates upp/nceppost in all other stack files. Please take a look both, and if you agreee merge it into Kyle's branch. Thanks!

climbfuji (approved):
Works on Cheyenne!

kgerheiser:
NCO and NCCMP use a CMake build which I don't think is affected.

aerorahul:
> NCO and NCCMP use a CMake build which I don't think is affected.

True, but.
in `build_nccmp.sh`
```
 cmake .. \
        -DBUILD_TESTS=OFF \
        -DCMAKE_BUILD_TYPE=Release \
        -DCMAKE_EXE_LINKER_FLAGS="$LDFLAGS" \
        -DCMAKE_INSTALL_PREFIX=$prefix \
        -DCMAKE_VERBOSE_MAKEFILE=ON \
        -DNETCDF_INC_DIR=$NETCDF_ROOT/include \
        -DNETCDF_LIB_PATH=$NETCDF_ROOT/lib/libnetcdf.a \
        -DWITH_NETCDF=${netcdf_src:-}
```

`LDFLAGS` is obtained as before.

aerorahul (approved):
looks good.

PR #206, opened 2021-04-12
Feature/1.2.0 updates

kgerheiser:
Fix DOWNLOAD_ONLY option by removing `git fetch`, using autotools build of PIO, and using `git reset --hard && git clean` to clean ESMF instead of re-downloading it.

Fix test options by swapping `make check` with `make test` for PNG and Jasper

Update ESMF to bs 47 which is what the weather model currently uses

Update PIO to 2.5.3 to fix a bug with `make check` 

Remove nceppost. I don't think this it is needed anymore.

### DIFF
--- a/libs/build_cdo.sh
+++ b/libs/build_cdo.sh
@@ -34,7 +34,7 @@ if $MODULES; then
   fi
 else
   prefix=${CDO_ROOT:-"/usr/local"}
-  enable_pnetcdf=$(nc-config --has-pnetcdf)
+  [[ ${DOWNLOAD_ONLY} =~ [yYtT] ]] || enable_pnetcdf=$(nc-config --has-pnetcdf)
 fi
 
 if [[ ! -z $mpi ]]; then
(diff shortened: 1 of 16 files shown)


arunchawla-NOAA:
Hi ESMF has done a formal 8.1.0 release does it make sense to just move to this? This is for @junwang-noaa with regards to ufs-weather-model

junwang-noaa:
Yes, it would be nice to have ESMF 8.1.0 release in hpc-stack/1.2.0 so that
the UFS model can use the release  version.

kgerheiser:
Ok, I'll change that to the 8.1.0 release then.

kgerheiser:
The Boost download site is down which is causing the build errors.

libs/build_cdo.sh
```diff
@@ -34,7 +34,7 @@ if $MODULES; then
   fi
 else
   prefix=${CDO_ROOT:-"/usr/local"}
-  enable_pnetcdf=$(nc-config --has-pnetcdf)
+  [[ ${DOWNLOAD_ONLY} =~ [yYtT] ]] || enable_pnetcdf=$(nc-config --has-pnetcdf)
```
aerorahul: Sorry, I don't understand what this change is for.

aerorahul (commented):
Just one comment in CDO.

libs/build_cdo.sh
kgerheiser: If DOWNLOAD_ONLY is turned on then `nc-config` isn't available.

aerorahul: Understood.

I think we can completely remove this line from here and place it right before we start building with
```
enable_pnetcdf=$(nc-config --has-pnetcdf)
```
It will be available there.

aerorahul (approved):
looks good.

PR #217, opened 2021-04-26
Add MADIS to hpc-stack

### LINKED ISSUE #215: add MADIS library to hpc-stack(?)
There's a library available from https://madis-data.ncep.noaa.gov/madis_api.shtml (current release version is v4.3) which is used by the GTS decoder suite on the current WCOSS.  For WCOSS2, the NCO SPA office are suggesting that this library should become part of hpc-stack and included as part of that installation on WCOSS2, rather than being installed as part of the GTS decoder suite as it is now on the current WCOSS.

The library was developed by [GSL](https://gsl.noaa.gov/) and is normally supported through madis-support@noaa.gov which is part of NCO.  GSL have assured me that the library is still supported, and in fact there will be a new v4.5 and possibly v4.6 available soon.  Even though madis-support is part of NCO, the SPA office have notified us that madis-support only officially supports installations on the IDP (see https://www.nco.ncep.noaa.gov/idsb/IDP_standards/ and https://www.nco.ncep.noaa.gov/pmb/primarysite/ for more information about the IDP), and they will not directly support installation on WCOSS2.  Instead, they're proposing that this library now become part of hpc-stack, meaning that any support issues will be directed back to EMC, and we'll in turn need to reach out ourselves directly to GSL for any library support should it become necessary.

The GTS decoder suite is the only application I'm aware of which uses this library on the WCOSS, and the decoders don't run anywhere else supported by hpc-stack, so as far as I know there's no need for this library to also be installed on hera, orion, gaea, or anywhere else that hpc-stack builds are currently done.  That said, GSL is part of NOAA (under OAR), so I presume there must be other applications somewhere/elsewhere in NOAA that use this library, though again I'm not personally aware of any.

Given the above, does it indeed make sense to bring this library under our hpc-stack umbrella?  If so, and if there is agreement to do this, then please let me know and I can submit a new issue (or just add to this one) with more details.

aerorahul:
This PR:
- closes #215 
- uses MADISs' primitive build and a couple of `sed` commands to make the build portable and compatible with hpc-stack (and standard) installation structure.

Some peculiarities of MADISs' primitive build:
- everything is hard-wired in `src/makefile`
- no `make install` step
- does not support parallel build `make -j 1`


aerorahul:
@kgerheiser Can I get a review on this?

jbathegit:
@aerorahul  Thanks very much for your efforts on this!

kgerheiser (approved):

PR #221, opened 2021-04-26
Add boostrap option for CMake

### LINKED ISSUE #220: Update CMake to 3.20.1
Building the weather model with external FMS requires CMake 3.19+ which will be installed with hpc-stack.

kgerheiser:
Platforms with CMake already installed can disable bootstrap (boostrap: NO) and use the exisiting CMake to build the new CMake and save compile time.

Fixes #220

### DIFF
--- a/config/stack_custom.yaml
+++ b/config/stack_custom.yaml
@@ -8,8 +8,9 @@ mpi:
   version: 4.0.1
 
 cmake:
-  build: NO
-  version: 3.17.2
+  build: YES
+  bootstrap: YES
+  version: 3.20.1
 
 jpeg:
   build: YES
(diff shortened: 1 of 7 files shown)


aerorahul (approved):
looks good.

kgerheiser:
CMake doesn't build on macOS using GCC because of some system header `Authorization.h`. Need to use Clang. I'm going to disable building it for macOS so that the CI can pass.

edwardhartnett:
Why are we building CMake instead of just requiring it?

kgerheiser:
We don't normally build it because it's available on our systems already. However, that is not always the case on all systems, or the version is too old. So, we have the option to build it. 

For the weather model to use the external build of FMS in hpc-stack it needs a newer version of CMake, and the quickest way to deploy that across our systems while we wait for system admins to install it is to use hpc-stack because that's already how we manage other libraries.

PR #223, opened 2021-04-26
Change Boost download URL

kgerheiser:
Bintray is down again today, and will be going offline on May 1. Use the new download location.


aerorahul (approved):
Looks good.

PR #225, opened 2021-04-27
Set FMS AVX2 flags for Cheyenne, patch FMS for GNU 9/10 on Cheyenne

climbfuji:
Here is a suggestion to apply patches to packages that will work not just for FMS.

Also set the AVX2 flags for compiling FMS. Not sure if these should be the default in `libs/build_fms.sh` or not, therefore just adding to the Cheyenne configs for now.

Suggestions welcome. Whatever solution we adopt should make it into v1.2.0 in my opinion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@
 *.swp
 *~
 *.DS_Store
-*.patch
 *.diff
 *.tar.gz
 *.zip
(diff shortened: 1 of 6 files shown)


aerorahul:
elegant.  
Don't forget to add the patch to the repo.

climbfuji:
> elegant.
> Don't forget to add the patch to the repo.

Thanks! Turns out I had to remove `*.patch` from `.gitignore`, too.

climbfuji:
Same here, can I get a review or even better an approval? Thanks!

aerorahul (approved):
Looks good.

PR #233, opened 2021-05-08
Enable python API's.

aerorahul:
This PR:
- extends hpc-stack to use and build python API's and virtual environments.
- builds bufr python API
- fixes bufr modulefile that points to library in `lib`.  bufr libraries are installed under `lib64` on Linux hosts.
- ensures nceplibs serial libraries don't use MPI.
- unifies `build_lib` interface to build nceplibs as well.  `build_nceplib` function is removed.


aerorahul:
closing temporarily ...

aerorahul:
@kgerheiser @climbfuji 
This is now ready for your review and input.

README.md
```diff
@@ -23,15 +23,15 @@ Code Manager: Kyle Gerheiser
...
-cheyenne | Dom Heinzeller
-gaea | Dom Heinzeller
-WCOSS-Cray | Hang-Lei
+Hera        | Kyle Gerheiser
+Jet         | Kyle Gerheiser
+Orion       | Hang-Lei
+WCOSS-Dell  | Hang-Lei
+WCOSS-Cray  | Hang-Lei
+Cheyenne    | Dom Heinzeller
+Gaea        | Dom Heinzeller
```
climbfuji, line 34: Gaea will probably change when we roll out v1.2.0

config/stack_mac.yaml
```diff
@@ -39,18 +39,18 @@ udunits:
 hdf5:
   build: YES
   version: 1.10.6
-  shared: NO
+  shared: YES
```
climbfuji, line 42: Why shared? Static libraries are better in my opinion. I need to change this file every time I build hpc-stack on my mac.

stack_helpers.sh
```diff
@@ -163,6 +163,7 @@ function set_no_modules_path() {
   local prefix=${PREFIX:-${HPC_OPT:-"/usr/local"}}
   export PATH=$prefix/bin${PATH:+:$PATH}
   export LD_LIBRARY_PATH=$prefix/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
+  export LD_LIBRARY_PATH=$prefix/lib64${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
```
climbfuji, line 166: Brute force ... lib64 may not even exist on the system, and LD_LIBRARY_PATH is definitely set because of the line above. But doesn't matter, it works just fine.

climbfuji (commented):
We will need to test this on gaea and cheyenne, probably need to load Python modules and add to the config files. Not happy with the shared libraries on macOS, but if there is a good reason I am happy to learn.

README.md
aerorahul: If one leaves out `HPC-PYTHON` in the `config.sh`, it will be ignored and no python api's will be built.
I left it underfined in  Gaea, Hera and WCOSS configs.
Hera is a mess.

config/stack_mac.yaml
aerorahul: I need to change it to shared, everytime I re-build on my macOS.
Most packages built on macOS are shared. 
I want to make sure the CI catches shared building.

stack_helpers.sh
aerorahul: You are right. If it doesn't, what does it matter?

config/stack_mac.yaml
aerorahul: Do you have a really really strong argument against using shared libraries?

climbfuji: `rpath` on macOS can mess up a lot, and there is always a danger that the wrong library gets picked if the runtime `LD_LIBRARY_PATH`/`DYLD_LIBRARY_PATH` doesn't match the one when the code is compiled. Or when homebrew decides that it wants to update a whole bunch of libraries accidentally (e.g. when the user wants to install something completely unrelated from homebrew). This doesn't happen with a static executable. But I won't insist on that, in the worst case I will change it every time or find another solution.

aerorahul: Homebrew does not look at what we install via hpc-stack. So unless one "removes" a package from homebrew that was used for building a software in hpc-stack, the rpath issue should not arise.  
I have been able to use different gnu compiler versions (gcc-9.1, 9.2, 9.3) with the same hpc-stack built with gcc-9.1 from homebrew with no issues. 
I have had more issues with Xcode updates breaking my software than homebrew updates.

climbfuji (approved):

kgerheiser:
Sorry, I didn't get to reviewing it. Looks good.

PR #235, opened 2021-05-10
purge modules in config.sh to ensure a clean environment for build

aerorahul:
A lot of developers/users like to load modules at login via bashrc.
This can pollute the environment and create issues that are very difficult to debug.

This PR:
- purges loaded modules before starting the build of the software stack
- ensures a clean build environment that is free from user/developer loaded modules in their bashrc or environment.


DusanJovic-NOAA (approved):

config/config_wcoss2.sh
```diff
@@ -35,6 +35,7 @@ export WGET="wget -nv"
 # define cc, FC and CC for CC, FC and CXX respectively.
 # cray-intel does not imply the native module in this case
 
+#module purge
```
climbfuji, line 38: Why is this commented out?

aerorahul: wcoss2 is not a regularly tested platform. 
Only a few people have access to it and it gets updates to it modules regularly.
I don't know what will break or what more will need to be loaded after a purge.
I was going to ask them to test with this and report back.
Easiest way to ask was to uncomment 1 line and run it.

climbfuji: Thanks, I was just curious.

climbfuji (approved):
Should do the same for Cheyenne, but not for Gaea (crays don't like that).

PR #238, opened 2021-05-11
Fix PIO build by using CMake

### LINKED ISSUE #211: Autotools build of PIO fails when using Intel compiler
**Describe the bug**
The PIO build was switched from CMake to autotools with #206 to fix the DOWNLOAD_ONLY option. Now I get this error `This module file was not generated by any release of this compiler`.

Which is caused by this

https://github.com/NCAR/ParallelIO/issues/1862

@edwardhartnett do you have a suggestion for a workaround? Might have to go back to the CMake for the time being.

kgerheiser:
PIO fails to build with Intel compilers using the Autoconf build.

Also fix link order in Madis

Fix #211

### DIFF
--- a/libs/build_pio.sh
+++ b/libs/build_pio.sh
@@ -77,7 +77,7 @@ mkdir -p build && cd build
 
 cmake ..\
   -DCMAKE_INSTALL_PREFIX=$prefix \
-  -DNetCDF_PATH=${NETCDF:-} \
+  -DNetCDF_PATH=${NETCDF_ROOT:-} \
   -DHDF5_PATH=${HDF5_ROOT:-} \
   -DCMAKE_VERBOSE_MAKEFILE=1 \
   $CMAKE_FLAGS
--- a/libs/build_pio.sh
+++ b/libs/build_pio.sh
@@ -53,53 +53,38 @@ else
   branch=pio$(echo $version | sed -e 's/\./_/g')
 fi
 
-URL="https://github.com/NCAR/ParallelIO/releases/download/$branch/${software}.tar.gz"
-[[ -f $software.tar.gz ]] || $WGET $URL
-
-tar -xf ${software}.tar.gz
-
-
+URL=https://github.com/NCAR/ParallelIO
+[[ -d $software ]] || git clone $URL $software
 [[ -d $software ]] && cd $software || ( echo "$software does not exist, ABORT!"; exit 1 )
+
+git checkout $branch
+
+# These repositories are used internally by PIO. Download them so DOWNLOAD_ONLY option works.
+[[ -d CMake_Fortran_utils ]] || git clone https://github.com/CESM-Development/CMake_Fortran_utils
+[[ -d genf90 ]] || git clone https://github.com/PARALLELIO/genf90.git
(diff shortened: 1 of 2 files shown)


climbfuji:
Let me pull this into my cheyenne testing code and see what happens ...

libs/build_pio.sh
```diff
@@ -53,53 +53,38 @@ else
...
-    TIMING_FLAGS="--enable-timing"
-else
-    TIMING_FLAGS=""
-fi
 
-export LDFLAGS="$PNETCDF_LDFLAGS $NETCDF_LDFLAGS $HDF5_LDFLAGS $AM_LDFLAGS"
-export LIBS="$HDF5_LIBS $EXTRA_LIBS"
+cmake ..\
+  -DCMAKE_INSTALL_PREFIX=$prefix \
+  -DNetCDF_PATH=${NETCDF:-} \
```
aerorahul: Shouldn't this be `NETCDF_ROOT`

libs/build_pio.sh
```diff
@@ -53,53 +53,38 @@ else
...
+[[ -d $software ]] || git clone $URL $software
+[[ -d $software ]] && cd $software || ( echo "$software does not exist, ABORT!"; exit 1 )
 
+git checkout $branch
 
-[[ -d $software ]] && cd $software || ( echo "$software does not exist, ABORT!"; exit 1 )
+# These repositories are used internally by PIO. Download them so DOWNLOAD_ONLY option works.
+[[ -d CMake_Fortran_utils ]] || git clone https://github.com/CESM-Development/CMake_Fortran_utils
+[[ -d genf90 ]] || git clone https://github.com/PARALLELIO/genf90.git
+CMAKE_FLAGS="-DUSER_CMAKE_MODULE_PATH=`pwd`/CMake_Fortran_utils -DGENF90_PATH=`pwd`/genf90"
```
aerorahul, line 65: not that I care whatever PIO is doing, but are these "required" by PIO?
If they are required, does anyone know why these are not git submodules?
If they are not required, what do we need them for?

aerorahul: if these are required for testing, better to put them in an if-block:
```
if [[ $MAKE_CHECK =~ [yYtT] ]]; then
...
fi
```

kgerheiser: Yes, they are required, if you want the ability to use `DOWNLOAD_ONLY`. Otherwise, they are downloaded by CMake at build time.

libs/build_pio.sh
kgerheiser: It can be. They point to the same thing.

aerorahul: In the modulefile, yes.  But when building without modules, the variable `NETCDF` is undefined, and only `NETCDF_ROOT` is defined.  I think.

kgerheiser: True, I didn't think of that.

aerorahul (approved):
looks good. thanks for the PIO and Madis fix.
Not sure why MADIS was not failing earlier.

climbfuji (approved):
Works on Cheyenne with Intel 19, GNU 9, GNU 10. Thanks for the bug fixes!

climbfuji:
@kgerheiser ready to merge?

PR #241, opened 2021-05-12
Configure Python environment for gaea and cheyenne, update nccmp to 1.8.9.0

### LINKED ISSUE #239: nccmp 1.8.7.0 build fails with GNU-10 - update nccmp to 1.8.9.0
**Describe the bug**
With GNU-10 (on Cheyenne, but that should be system independent), building `nccmp` fails with
```
mpicc -I../.. -I/glade/scratch/heinzell/hpc-stack/source/hpc-stack-develop-20210511-gnu10/pkg/netcdf-c-4.7.4/include -I/glade/scratch/heinzell/hpc-stack/source/hpc-stack-develop-20210511-gnu10/pkg/netcdf-c-4.7.4/libsrc4 -fPIC   -L/glade/scratch/heinzell/hpc-stack/source/install/gnu-10.1.0/mpt-2.22/netcdf/4.7.4/lib -L/glade/scratch/heinzell/hpc-stack/source/install/gnu-10.1.0/mpt-2.22/hdf5/1.10.6/lib   -L/glade/scratch/heinzell/hpc-stack/source/install/gnu-10.1.0/zlib/1.2.11/lib  -o nccmp nccmp-log.o nccmp-nccmp.o nccmp-xmalloc.o nccmp-nccmp_buffer.o nccmp-nccmp_data.o nccmp-nccmp_darray.o nccmp-nccmp_darray_sort.o nccmp-nccmp_dim.o nccmp-nccmp_format.o nccmp-nccmp_group.o nccmp-nccmp_io.o nccmp-nccmp_metadata.o nccmp-nccmp_nc_type.o nccmp-nccmp_ncinfo.o nccmp-nccmp_nest.o nccmp-nccmp_odometer.o nccmp-nccmp_opt.o nccmp-nccmp_state.o nccmp-nccmp_stats.o nccmp-nccmp_stats_print.o nccmp-nccmp_strlist.o nccmp-nccmp_utils.o nccmp-nccmp_user_type.o nccmp-nccmp_var.o   -lnetcdf -lm  -lnetcdf -lm  -lnetcdf -lhdf5_hl -lhdf5  -lz -ldl -lm
/usr/bin/ld: nccmp-nccmp_group.o:(.bss+0x0): multiple definition of `mutex_print'; nccmp-nccmp_data.o:(.bss+0x0): first defined here
/usr/bin/ld: nccmp-nccmp_metadata.o:(.bss+0x0): multiple definition of `mutex_print'; nccmp-nccmp_data.o:(.bss+0x0): first defined here
collect2: error: ld returned 1 exit status
Makefile:393: recipe for target 'nccmp' failed
make[2]: *** [nccmp] Error 1
make[2]: Leaving directory '/glade/scratch/heinzell/hpc-stack/source/hpc-stack-develop-20210511-gnu10/pkg/nccmp-1.8.7.0/build/src'
Makefile:380: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/glade/scratch/heinzell/hpc-stack/source/hpc-stack-develop-20210511-gnu10/pkg/nccmp-1.8.7.0/build'
Makefile:321: recipe for target 'all' failed
make: *** [all] Error 2
BUILD FAIL!  Lib: nccmp Error:2
```
This is not a bug in `hpc-stack`, but in `nccmp`. The solution is to update `nccmp` to the latest (4 months old) tag 1.8.9.0, which builds fine with Intel 19, GNU 9 and GNU 10 on Cheyenne.

climbfuji:
This PR
- configures the Python environment for Cheyenne (use dummy environment tto turn off) and Gaea
- adds several python meta modules (contributed by @aerorahul)
- update `stack_gaea.yaml` (brings it closer to `stack_noaa.yaml`)
- updates `nccmp` from 1.8.7.0 to 1.8.9.0 in all stack yaml files
- skip asking whether to overwritee module files if `OVERWRITE` is true
- includes a few cosmetic changes (stdout) in `setup_modules.sh`

Fixes https://github.com/NOAA-EMC/hpc-stack/issues/239

This has been tested on Cheyenne with Intel 19, GNU 9, GNU 10, and on Gaea with Intel 18.


config/config_cheyenne_gnu-10.sh
```diff
@@ -30,6 +31,11 @@ module load ncarenv/1.3
 # Load these basic modules for Cheyenne
 module load cmake/3.18.2
 
+# Load Python environment
+module use /glade/p/ral/jntp/tools/miniconda3/4.8.3/modulefiles
+module load miniconda3/4.8.3
+conda activate ufs-weather-model
```
aerorahul: If one does this here, one would be putting the `PATH` and `PYTHONPATH` set by these modules in the default environment of every single build script.

How does activating `ufs-weather-model` conda env translate to other/generic applications?
Is that even necessary at build time?

I would suggest doing something like this:
```
export HPC_PYTHON="miniconda3/4.8.3"

# Load Python env.
module use /glade/p/ral/jntp/tools/miniconda3/4.8.3/modulefiles
```

Add a directory with a template modulefile `modulefiles/core/hpc-miniconda3/miniconda3.lua` following the example for `python/python.lua`

climbfuji: I removed the `conda activate calls`, they were not needed. Also added missing GNU-10 compatibility flags for madis. I am not going to dive into these module files, the current solution works and doesn't break anything. But I did notice some copy and paste errors in the module files when I briefly looked at them.

climbfuji:
I got rid of the messy commit history by doing a `git rebase`, now it looks clean.

config/config_cheyenne_gnu-10.sh
```diff
@@ -30,9 +31,14 @@ module load ncarenv/1.3
 # Load these basic modules for Cheyenne
 module load cmake/3.18.2
 
+# Load Python environment
+module use /glade/p/ral/jntp/tools/miniconda3/4.8.3/modulefiles
+module load miniconda3/4.8.3
```
aerorahul: This module should not be loaded in config; just like we don't load the compiler in the config.
If for e.g. netcdf is built in miniconda, it can cause all kinds of headaches when building the software.

kgerheiser (approved):

aerorahul (changes requested):
Should not load miniconda in the config.sh.
This should be loaded where needed as is the case of compiler and mpi

config/config_cheyenne_gnu-10.sh
aerorahul: Can you point out the copy/paste errors.  They should be fixed.

It works does not mean that is an acceptable solution.

config/config_cheyenne_gnu-10.sh
climbfuji: Ok, this explanation makes sense to me. Note that the config files for Cheyenne and Gaea do (and always did) contain module load/switch statements for the compiler (and MPI). I don't think it works on Cheyenne otherwise because of the way CISL designed the ncarcompilers and ncarenv modules.

config/config_cheyenne_gnu-10.sh
climbfuji: One I remember is in `modulefiles/core/python/python.lua`:
```
...
whatis("Description: GNU Compiler Family")
```
not sure if there are others.

aerorahul: legit copy paste error and me being sloppy.  Thanks!
thankfully it is not a fatal error and will show false description if someone did `module show python`

config/config_cheyenne_gnu-10.sh
climbfuji: I was able to build everything with the changes I just pushed for the cheyenne and gaea config*.sh files. I didn't get the hpc python modules, presumably because I did not create the miniconda meta modules. I really don't know what I would be doing there, can I ask you to push a miniconda module to my branch?

climbfuji:
Thanks a lot for creating the miniconda modules, @aerorahul! I will test those today, as well as the updated JEDI packages, and resolve the conflict for `stack_gaea.yaml`.

climbfuji:
> Thanks a lot for creating the miniconda modules, @aerorahul! I will test those today, as well as the updated JEDI packages, and resolve the conflict for `stack_gaea.yaml`.

There seems to be one issue, and that is that I need to run
```
module use /glade/p/ral/jntp/tools/miniconda3/modulefiles
```
before I can run
```
module load hpc-miniconda3/4.8.3
```
otherwise the latter command fails with
```
heinzell@cheyenne3:~> module load   hpc-miniconda3/4.8.3
Lmod has detected the following error:  Cannot load module "hpc-miniconda3/4.8.3" without these module(s) loaded:
   miniconda3/4.8.3

While processing the following module(s):
    Module fullname       Module Filename
    ---------------       ---------------
    hpc-miniconda3/4.8.3  /glade/scratch/heinzell/hpc-stack/install/modulefiles/core/hpc-miniconda3/4.8.3.lua
```

I don't think the hpc-stack metamodule can find out where to find tthe miniconda3 module, hence having to do `module use` first is something we might have to live with, but the metamodule should complain about not being able to find miniconda3, not that it has to be loaded first.

climbfuji:
@aerorahul good news, thanks to yesterday's jedi software update, we could get rid of `stack_gaea.yaml` - we can now use `stack_noaa.yaml`.

climbfuji:
> > Thanks a lot for creating the miniconda modules, @aerorahul! I will test those today, as well as the updated JEDI packages, and resolve the conflict for `stack_gaea.yaml`.
> 
> There seems to be one issue, and that is that I need to run
> 
> ```
> module use /glade/p/ral/jntp/tools/miniconda3/modulefiles
> ```
> 
> before I can run
> 
> ```
> module load hpc-miniconda3/4.8.3
> ```
> 
> otherwise the latter command fails with
> 
> ```
> heinzell@cheyenne3:~> module load   hpc-miniconda3/4.8.3
> Lmod has detected the following error:  Cannot load module "hpc-miniconda3/4.8.3" without these module(s) loaded:
>    miniconda3/4.8.3
> 
> While processing the following module(s):
>     Module fullname       Module Filename
>     ---------------       ---------------
>     hpc-miniconda3/4.8.3  /glade/scratch/heinzell/hpc-stack/install/modulefiles/core/hpc-miniconda3/4.8.3.lua
> ```
> 
> I don't think the hpc-stack metamodule can find out where to find tthe miniconda3 module, hence having to do `module use` first is something we might have to live with, but the metamodule should complain about not being able to find miniconda3, not that it has to be loaded first.

@aerorahul any chance to fix the error message (not the error, that seems to be impossible to me)? Then this PR is ready, it compiled everything successfully on Cheyenne with Intel, GNU9, GNU10, and Gaea with Intel.

config/config_gaea.sh
```diff
@@ -3,12 +3,13 @@
...
 export HPC_COMPILER="intel/18.0.6.288"
 export HPC_MPI="cray-mpich/7.7.11"
+export HPC_PYTHON="cray-python/3.7.3.2"
 
 # Build options
 export USE_SUDO=N
 export PKGDIR=pkg
 export LOGDIR=log
-export OVERWRITE=N
+export OVERWRITE=Y
```
kgerheiser, line 12: Is this meant to be on?

climbfuji: As long as I maintain the stack on gaea, yes please. It's on for Cheyenne, too.

aerorahul (approved):
looks good. Thanks for the work.

PR #245, opened 2021-05-21
add customized config and stack files for community GSI/EnKF

guoqing-noaa:
Community GSI/EnKF users can use config_comgsi.sh and stack_comgsi.yaml to build only those libraries required to compile regional GSI/EnKF.

### DIFF
--- a/config/config_comgsi.sh
+++ b/config/config_comgsi.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Compiler/MPI combination
+export HPC_COMPILER=${HPC_COMPILER:-"intel/18.0.5"}
+export HPC_MPI=${HPC_MPI:-"impi/2018.4.274"}
+export HPC_PYTHON=${HPC_PYTHON:-"python/3.9.4"}
+
+# Build options
+export USE_SUDO=N
+export PKGDIR=pkg
+export LOGDIR=log
+export OVERWRITE=Y
+export NTHREADS=4
+export   MAKE_CHECK=N
+export MAKE_VERBOSE=N
+export   MAKE_CLEAN=N
+export DOWNLOAD_ONLY=N
+export STACK_EXIT_ON_FAIL=Y
+export WGET="wget -nv"
--- a/config/stack_comgsi.yaml
+++ b/config/stack_comgsi.yaml
@@ -184,8 +184,8 @@ wrf_io:
 
 bufr:
   build: YES
-  version: bufr_v11.4.0
-  install_as: 11.4.0
+  version: bufr_v11.5.0
+  install_as: 11.5.0
 
 wgrib2:
   build: NO
--- a/config/stack_comgsi.yaml
+++ b/config/stack_comgsi.yaml
@@ -1,0 +1,277 @@
+gnu:
+  build: NO
+  version: 9.3.0
+
+mpi:
+  build: NO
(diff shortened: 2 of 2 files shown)


edwardhartnett:
Do these packages have a cmake build system?

We need to get a cmake build system for all packages in hpc-stack. If there is not currently one, we need to come up with a plan to transition to a cmake build.

We cannot support home-rolled build scripts for 50+ libraries and packages, so we require a standard build system.

kgerheiser:
@edwardhartnett this doesn't really have anything to do with a build system. It's the same packages that are already here, but with some disabled because GSI doesn't need them.

edwardhartnett:
OK never mind...

config/config_comgsi.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Compiler/MPI combination
+export HPC_COMPILER=${HPC_COMPILER:-"intel/18.0.5"}
```
kgerheiser, line 4: What machine does this config refer to?

guoqing-noaa: That value is for Cheyenne but these three HPC_* variables are just place holders.  Community users needs to change them or set corresponding environments for hpc-stack to work (community users spread around the world, not limited to NOAA and NCAR machines). Thanks!

kgerheiser: Do we need this specific config then? I'm fine with the `stack_comgsi`, but I'm not sure this config adds anything.

guoqing-noaa: I would hope to add them. But I am okay if it is decided to not include them and if so, community users will need to get a copy from the comgsi/hpc-stack fork (not as official as from NOAA-EMC/hpc-stack).

kgerheiser: Maybe a different name then? It's not clear to me that this is meant to be used on Cheyenne, and none of the other settings are unique. You can't just run use this config on any machine. Or is GSI development usually done on Cheyenne?

guoqing-noaa: Do you have any suggestions on "a different name"?  Yes, I mainly develop and test community GSI on Cheyenne.

kgerheiser: How about `config_cheyenne_GSI?`

guoqing-noaa: Ah, I think that may not be appropriate. These two files are NOT used on Cheyenne solely.  I also used it on Orion to build libraries for regional GSI/EnKF ( I only need to export HPC_COMPILER and HPC_MPI variables). Users also use them to build their comgsi-flavor hpc-stack, for example on Stampede of TACC, on supercomputers at CSU, etc. I got what you mean. You'd like to see this file is specific to one particular machine. But my intention is to get ready a file specifically customized for building community GSI (mostly for regional applications). If this goal conflicts with some hpc-stack policies, it is fine to not include these two files. Thanks!

kgerheiser: I'm convinced with your explanation.

kgerheiser (approved):

config/config_comgsi.sh
guoqing-noaa: Thanks!

guoqing-noaa:
Could someone merge this PR as it was already approved by @kgerheiser ?  @Hang-Lei-NOAA

PR #253, opened 2021-06-24
Correctly append OpenMPI to PATH if cached

kgerheiser:
OpenMPI was only being appended to PATH when being built, but if it were cached the build step is skipped and it wouldn't be available.

### DIFF
--- a/.github/workflows/build_ubuntu.yaml
+++ b/.github/workflows/build_ubuntu.yaml
@@ -23,7 +23,6 @@ jobs:
     - name: build-openmpi
       if: steps.cache-mpi.outputs.cache-hit != 'true'
       run: |
-        echo "$HOME/mpi/bin" >> $GITHUB_PATH
         if [[ ${{ matrix.mpi }} == "openmpi" ]]; then
           wget https://download.open-mpi.org/release/open-mpi/v4.1/openmpi-4.1.1.tar.gz &> /dev/null
           tar -xzf openmpi-4.1.1.tar.gz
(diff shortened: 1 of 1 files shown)


Hang-Lei-NOAA (approved):

kgerheiser:
Combined with #256

.github/workflows/build_ubuntu.yaml
```diff
@@ -41,6 +40,9 @@ jobs:
 
     - name: Install dependencies
       run: |
+        if [[ ${{ matrix.mpi }} == "openmpi" ]]; then
+          echo "$HOME/mpi/bin" >> $GITHUB_PATH
+        fi
```
aerorahul, line 45: looks right. 
I forget now, is there no openmpi packages on Ubuntu?

kgerheiser: There is, but when using that you get undefined references to functions from the MPI C++ library (from ESMF) when linking executables using the MPI Fortran wrapper. It doesn't link to `-lmpi_cxx`. If you build it yourself you don't get that.

I've seen this in other repos (UFS_UTILS). I think it's related to how Ubuntu builds the MPI binaries, but I'm not exactly sure.

aerorahul: Thanks @kgerheiser 
[This](https://www.rosettacommons.org/node/1850) post seems to suggest that there is a solution.  May be worth a try in  the future and save us some compile time in the action.

kgerheiser: It's cached so it doesn't build it every time, but I will look at that.

PR #254, opened 2021-06-24
Remove CMAKE_PREFIX_PATH from yafyaml build

kgerheiser:
This variable is not needed. I get an unbound variable when building on Orion.


aerorahul (approved):
yep. looks right.

kgerheiser:
#253 is needed first

rmontuoro (approved):
Agree. `CMAKE_PREFIX_PATH` is no longer required.

kgerheiser:
Combing with other PRs in #256

PR #260, opened 2021-07-01
Add Intel build with LMod

kgerheiser:
It will be useful to have an Intel build and a module build.


.github/workflows/build_intel.yaml
```diff
@@ -0,0 +1,69 @@
+name: Build Intel
```
aerorahul, line 1: This is excellent.

aerorahul (approved):
This is super helpful to have.
Thanks @kgerheiser

PR #261, opened 2021-07-01
For v1.2.0: Update Python on Hera

climbfuji:
Rename `config/config_hera.sh` to `config/config_hera_intel.sh` and update Python version used, add `config/config_hera_gnu.sh`.

Update cmake on hera.


config/config_hera_gnu.sh
```diff
@@ -0,0 +1,27 @@
...
+export DOWNLOAD_ONLY=N
+export STACK_EXIT_ON_FAIL=Y
+export WGET="wget -nv"
+
+# Load these basic modules for Hera
+module purge
+module load cmake/3.16.1
+
+# To find miniconda3/3.7.3
+module use /scratch1/NCEPDEV/nems/emc.nemspara/soft/modulefiles
```
aerorahul, line 26: Is there something wrong with:
```
module use /contrib/miniconda3/modulefiles
```

It provides a miniconda3 modulefile for version 4.5.12?

I have been using that for creating a virtual env. for JEDI work.

climbfuji: There is no virtual environment in this setup, everything is part of the base environment of this installation. Making the change as you suggested will not have all the dependencies. If someone wants to change that, then that's great, but it won't work right away.

UPDATE: Let me be clear, it will be fine for hpc-stack itself, but not for the ufs-weather-model. It would be good if we could be consistent in which version of Python is used in the hpc-stack install and the ufs-weather-model.

climbfuji:
We should update cmake ... let me do that. cmake 3.20.1 is a default module on hera nowadays

config/config_hera_gnu.sh
aerorahul: And that is not appropriate.
Is emc.nemspara going to provide the python packages for all developers?
The correct way to do this is to use base python, and create python virtual environments for the applications with the python modules (and their respective versions) that the applications require.

climbfuji: Please go ahead and update this with the next ufs-weather-model PR. The problem has been discussed for days and held back the GOCART PR, nobody was willing to take action.

Note that using the conda init, activate and deactivate functionality in many cases does not work, one still has to manually set PYTHONPATH, PATH, LD_LIBRARY_PATH or create modulefiles to do that.

aerorahul: There is no `conda init`, `conda activate`, `conda deactivate` in the case of python virtual environments. 
These "new" GOCART dependencies came out of nowhere.  There was a previous PR to hpc-stack that added GOCART dependencies.

If you give me a list of which python packages that are required for the UFS weather model, I can try and create a PR with those as a python virtual env. and its associated modulefile.

climbfuji: That would be great. What we need is Python3 pandas and netCDF4 (these were the ones that I had to install), and ecflow (unless we pick that from somewhere else). That's all I think.

climbfuji: We could combine switching to the virtual env with rolling out hpc-stack v.1.2.0 that then uses the same miniconda3. Thanks for helping with this!

aerorahul:
@climbfuji 
I have a branch `feature/updates-to-pyvenv` that contains what you need for the ufs-weather-model PR's python dependencies.  I will issue a PR for this as it has work I needed to do for JEDI python packages.

This update does not provide `ecflow`, so we will have to pick that from elsewhere.
Please let me know if anything is missing or not working as expected.

The procedure for Hera is exactly the same.  The packages need to be downloaded `(DOWNLOAD_ONLY=YES)` some place else and moved to the `pkg` dir on Hera.
This is also true for the python dependencies.
However, in the case of these python dependencies, they should be downloaded on a Linux machine.  If you download the python packages on a macOS, `pip` will find macOS wheels.

climbfuji:
> @climbfuji
> I have a branch `feature/updates-to-pyvenv` that contains what you need for the ufs-weather-model PR's python dependencies. I will issue a PR for this as it has work I needed to do for JEDI python packages.
> 
> This update does not provide `ecflow`, so we will have to pick that from elsewhere.
> Please let me know if anything is missing or not working as expected.
> 
> The procedure for Hera is exactly the same. The packages need to be downloaded `(DOWNLOAD_ONLY=YES)` some place else and moved to the `pkg` dir on Hera.
> This is also true for the python dependencies.
> However, in the case of these python dependencies, they should be downloaded on a Linux machine. If you download the python packages on a macOS, `pip` will find macOS wheels.

Sounds good! We need `ecflow`, thus the PR will have to provide that unless it is available separately on all platforms that we are concerned with. The easiest is to install it via conda into the virtual environment.

aerorahul:
> > @climbfuji
> > I have a branch `feature/updates-to-pyvenv` that contains what you need for the ufs-weather-model PR's python dependencies. I will issue a PR for this as it has work I needed to do for JEDI python packages.
> > This update does not provide `ecflow`, so we will have to pick that from elsewhere.
> > Please let me know if anything is missing or not working as expected.
> > The procedure for Hera is exactly the same. The packages need to be downloaded `(DOWNLOAD_ONLY=YES)` some place else and moved to the `pkg` dir on Hera.
> > This is also true for the python dependencies.
> > However, in the case of these python dependencies, they should be downloaded on a Linux machine. If you download the python packages on a macOS, `pip` will find macOS wheels.
> 
> Sounds good! We need `ecflow`, thus the PR will have to provide that unless it is available separately on all platforms that we are concerned with. The easiest is to install it via conda into the virtual environment.

I am exploring how to install ecflow via pip installation.  conda is not available on WCOSS (but WCOSS provides ecflow separately).  Hera is about to provide ecflow via a modulefile as well.

edwardhartnett:
Can we have an issue describing this work please?

Before a PR, we need an issue. This is a vital part of our documentation and can't be skipped...

climbfuji:
I think this PR is no longer needed, correct? Except for the cmake update on hera, maybe @kgerheiser can include it in his "update config" PR?

kgerheiser:
@aerorahul added Cmake/3.20 to the Hera config in his PR

climbfuji:
> @aerorahul added Cmake/3.20 to the Hera config in his PR

Great, thanks, then this PR can be closed.

aerorahul:
I am testing miniconda ability in the stack in this branch:
https://github.com/NOAA-EMC/hpc-stack/tree/feature/conda
It is an alternative/enhancements to the python virtual envs somewhat related to the issue #264

PR #262, opened 2021-07-02
Updates to python virtual env.

aerorahul:
This PR:
* replaces `intelpython` with `miniconda3` on Hera for Python.
* updates `cmake` on Hera to 3.20.1
* improves `build_pyvenv.sh`
** provides ability to `DOWNLOAD_ONLY` for python package wheels.  The wheels should be downloaded on "similar" machines.  For e.g. downloading wheels on macOS for use on Linux OS will not work.
* generalizes how python virtual environment modules are created.
** do not need to create a modulefile for each individual python virtual environment.
* adds a recipe for python dependencies in the ufs-weather-model.

### DIFF
--- a/config/config_hera.sh
+++ b/config/config_hera.sh
@@ -3,7 +3,7 @@
 # Compiler/MPI combination
 export HPC_COMPILER="intel/18.0.5.274"
 export HPC_MPI="impi/2018.0.4"
-export HPC_PYTHON="intelpython/3.6.8"
+export HPC_PYTHON="miniconda3/4.5.12"
 
 # Build options
 export USE_SUDO=N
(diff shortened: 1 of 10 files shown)


kgerheiser (approved):

kgerheiser:
What do `hofx` and `ufswm` mean?

aerorahul:
> What do `hofx` and `ufswm` mean?

They are both python virtual envs and I should update the `README.md` to make a note of this.  I will do that in the next PR:
- `hofx` is used by the Data assimilation group for their HofX application.  It is still under development.
- `ufswm` is used for compiling the UFS weather model.  The recent `GOCART` addition is requiring `pandas` for preprocessing of Fortran code.  Also added is netCDF used in regression testing.

edwardhartnett:
Can we have one or more issues describing this work please?

Issue #265, opened 2021-07-07
Incorrect library locations in g2 module file


DavidHuber-NOAA:
**Describe the bug**
The module file for g2 creates the environmental variables G2_LIB4 and G2_LIBd pointing to $base/lib/libg2_4.a and $base/lib/libg2_d.a, respectively.  However, these libraries actually install to $base/lib**64**.

**To Reproduce**
This can be found on orion
```
module use /apps/contrib/NCEP/libs/hpc-stack/modulefiles/stack
module load hpc/1.1.0 hpc-intel/2018.4 g2/3.4.2
env | grep G2_LIB4
>G2_LIB4=/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/g2/3.4.2/lib/libg2_4.a
ls $G2_LIB4
>ls: cannot access /apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/g2/3.4.2/lib/libg2_4.a: No such file or directory
ls /apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/g2/3.4.2/lib64/libg2_4.a
>/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/g2/3.4.2/lib64/libg2_4.a
```

**Expected behavior**
The environmental variables would point to the correct locations.

**System:**
Orion and S4

**Additional context**
This could be fixed by changing the line [22-23 of g2.lua](https://github.com/NOAA-EMC/hpc-stack/blob/19bdd88c14b9e76324e7b27af7b648e1341bb228/modulefiles/compiler/compilerName/compilerVersion/g2/g2.lua#L22-L23).

kgerheiser:
This is caused by CMake using `CMAKE_INSTALL_LIBDIR`, which depending on platform can be either lib or lib64. 

What we need is to be able to dynamically detect lib or lib64. We have an existing solution which processes the module before installing it. In the module we need to change `lib/*` to `{CMAKE_INSTALL_LIBDIR}/*`. We were using a version of g2 which wasn't using `CMAKE_INSTALL_LIBDIR`, but the version of g2 was updated and that version uses CMAKE_INSTALL_LIBDIR.

I can manually fix the existing discrepancy on Orion.

edwardhartnett:
Was there an issue for this?

aerorahul:
PR #266 resolved this issue.

PR #266, opened 2021-07-07
Correct the location of the g2 library paths

DavidHuber-NOAA:
Corrected the location of the `$G2_LIB4` and `$G2_LIBd` environmental variables (for libg2_4.a and libg2_d.a, respectively) in the g2.lua module file template to be g2/lib64 instead of g2/lib.


kgerheiser:
This is not correct because it can be either `lib` or `lib64` depending on the system. If you change it to `${CMAKE_INSTALL_LIBDIR}` then I can approve this PR.

kgerheiser:
See here for an example: https://github.com/NOAA-EMC/hpc-stack/blob/develop/modulefiles/compiler/compilerName/compilerVersion/wgrib2/wgrib2.lua

DavidHuber-NOAA:
I gave this a quick test build on S4 and it sets the variables as expected.  Thanks!

kgerheiser (approved):

edwardhartnett:
Is there an issue for this? If not, please add one.

kgerheiser:
https://github.com/NOAA-EMC/hpc-stack/issues/265

PR #268, opened 2021-07-08
Port hpc-stack to S4

### LINKED ISSUE #258: Port to S4
To support the port of the global workflow to SSEC's S4 cluster, hpc-stack should be ported and installed on that system.

DavidHuber-NOAA:
Added `config_s4.sh` and the new module file `hpc-miniconda.lua`, modeled after `hpc-miniconda3.lua`, to address #258.

The official install of hpc-stack on S4 is located on `/data/prod/hpc-stack` and can be accessed with `module use /data/prod/hpc-stack/modulefiles/stack`.

closes #258

### DIFF
--- a/config/config_s4.sh
+++ b/config/config_s4.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# Compiler/MPI combination
+export HPC_COMPILER="intel/18.0.3"
+export HPC_MPI="impi/18.0.3"
+export HPC_PYTHON="miniconda/3.8-s4"
+
+# Build options
+export USE_SUDO=N
+export PKGDIR=pkg
+export LOGDIR=log
+export OVERWRITE=N
+export NTHREADS=8
+export   MAKE_CHECK=N
+export MAKE_VERBOSE=N
+export   MAKE_CLEAN=N
+export DOWNLOAD_ONLY=N
+export STACK_EXIT_ON_FAIL=Y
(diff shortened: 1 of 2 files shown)


kgerheiser (approved):

kgerheiser:
Looks good. Thanks for the port.

config/config_s4.sh
```diff
@@ -0,0 +1,24 @@
...
+export MAKE_VERBOSE=N
+export   MAKE_CLEAN=N
+export DOWNLOAD_ONLY=N
+export STACK_EXIT_ON_FAIL=Y
+export WGET="wget -nv"
+
+# Load these basic modules for S4
+module purge
+module load license_intel/S4
+module load git/2.30.0
```
aerorahul, line 24: I think you will need to add a line like this:
```
module use /path/to/miniconda/3.8-s4/modulefile
```
to be able to load the miniconda/3.8-s4  modulefile unless it is being installed via hpc-stack, which I don't think we do, yet.

aerorahul (approved):
Looks good.
I made one comment in `config_s4.sh` to check if that is what you mean.

config/config_s4.sh
DavidHuber-NOAA: On S4, the path to miniconda/3.8-s4 is in `$MODULEPATH` by default, so it's good to go.  Thanks for checking.

Issue #270, opened 2021-07-12
Update configs to match default installations


kgerheiser:
Some of the configs don't have correct defaults. For example, Orion has `intel/2020` enabled by default, but we use `2018.4`. Or there's no `config_jet.sh`.

Some are also missing the AVX flags for FMS, which should be enabled on Hera, Orion, and WCOSS-Dell.

kgerheiser:
Fixed with #272

PR #272, opened 2021-07-12
Update configs

### LINKED ISSUE #271: Fix Cron job tests
The weather model has updated its module structure. Update the logic to work with `ufs_common`, `ufs_common_debug`.

kgerheiser:
Create a `config_jet.sh`

Update `config_orion.sh` to use correct compilers by default

Add vectorization flags to configs that are missing them.

Update the Cron testing job for the updated modulefile structure that the ufs weather model uses

Fix #271 #270

### DIFF
--- a/config/config_hera.sh
+++ b/config/config_hera.sh
@@ -22,3 +22,7 @@ export WGET="wget -nv"
 module purge
 module load cmake/3.20.1
 module use /contrib/miniconda3/modulefiles
+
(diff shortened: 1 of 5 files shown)


aerorahul (approved):
Looks good.  Thanks @kgerheiser

climbfuji (approved):
Thanks, Kyle.

PR #277, opened 2021-07-15
Build miniconda and conda environments.

### LINKED ISSUE #276: Build MIniconda and Conda environments capability
No necessarily, but current solutions are limited.
Python virtual environments are the way to isolate and protect from updates in the underlying python packages.
Python virtual environment (`venv`) is implemented in the hpc-stack.  This poses a limitation on machines (`Hera`) where the python distribution is handled by Miniconda.  One can also use `venv` with `pip`, but that requires downloading wheels and transferring them locally.
This PR builds Miniconda and allows for building conda environments in addition to python virtual environments.

Use Miniconda as the primary provider for python.

aerorahul:
Fixes #276 

This PR:
- adds the miniconda package as an option to be built within the stack
- adds the option to build conda environments.

### DIFF
--- a/libs/build_miniconda3.sh
+++ b/libs/build_miniconda3.sh
@@ -44,6 +44,7 @@ URL="https://repo.anaconda.com/miniconda/$installer"
 $SUDO bash $software/$installer -b -p $prefix -s
 set +x
 echo "sourcing conda.sh"
+PS1=
 source $prefix/etc/profile.d/conda.sh
 echo "setting conda default threads to 4"
 conda config --system --set default_threads 4
--- a/libs/build_miniconda3.sh
+++ b/libs/build_miniconda3.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# Download and install Miniconda3
+# https://docs.conda.io/projects/conda/en/latest/user-guide/getting-started.html
+
+set -eux
+
+name="miniconda3"
+version=${2:-${STACK_miniconda3_version:-"latest"}}
+
+if $MODULES; then
+  prefix="${PREFIX:-"/opt/modules"}/core/$name/$version"
+  if [[ -d $prefix ]]; then
+    [[ $OVERWRITE =~ [yYtT] ]] && ( echo "WARNING: $prefix EXISTS: OVERWRITING!"; $SUDO rm -rf $prefix; $SUDO mkdir $prefix ) \
+                               || ( echo "WARNING: $prefix EXISTS, SKIPPING"; exit 1 )
+  fi
+else
+  prefix=${MINICONDA3_ROOT:-"/usr/local"}
+fi
+
+cd ${HPC_STACK_ROOT}/${PKGDIR:-"pkg"}
+
+host=$(uname -s)
+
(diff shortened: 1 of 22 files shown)


kgerheiser (approved):

pyvenv/ufswm.txt
```diff
@@ -3,6 +3,5 @@ Cython >= 0.29.21
 numpy >= 1.19.4
 pandas >= 1.1.5
 python-dateutil >= 2.8.1
-h5netcdf >= 0.8.1
```
climbfuji, line 6: I guess you are leaving it up to netcdf4 to figure out its dependencies.

climbfuji (approved):

pyvenv/ufswm.txt
aerorahul: yes. That one is tricky because of incompatibilities on which version of netcdf-python is used.  Or atleast that is what I found.

libs/build_miniconda3.sh
```diff
@@ -0,0 +1,58 @@
...
+installer="Miniconda3-${pkg_version}-${os}-x86_64.sh"
+URL="https://repo.anaconda.com/miniconda/$installer"
+
+[[ -d $software ]] || ( mkdir -p $software; $WGET $URL -O $software/$installer )
+[[ ${DOWNLOAD_ONLY} =~ [yYtT] ]] && exit 0
+
+$SUDO bash $software/$installer -b -p $prefix -s
+set +x
+echo "sourcing conda.sh"
+source $prefix/etc/profile.d/conda.sh
```
kgerheiser, line 48: I get an unbounb variable `PS1` when sourcing this. ~~Should add a `set +eu` around it.~~


~~set +eu~~
~~ source $prefix/~etc/profile.d/conda.sh ~~
~~set -eu~~

Edit that didn't work.

kgerheiser (changes requested):

libs/build_miniconda3.sh
climbfuji: > I get an unbounb variable `PS1` when sourcing this. Should add a `set +eu` around it.
> 
> ```
> set +eu
> source $prefix/etc/profile.d/conda.sh
> set -eu
> ```

This is an known bug in the conda scripts, I think. If your suggestion works, then great.

kgerheiser: That didn't work, but `export PS1=""` worked. Not sure if there's a better way.

aerorahul: instead of compromising on error trapping, what if we do:
```
PS1=
source $prefix/etc/profile.d/conda.sh
```

kgerheiser: Yes, that works.

aerorahul:
```suggestion
PS1=
source $prefix/etc/profile.d/conda.sh
```
@kgerheiser Can I commit this change?

kgerheiser: Yes

kgerheiser (approved):

kgerheiser:
That commit suggestion thing is pretty nifty

PR #293, opened 2021-08-07
Wcoss2 update

### LINKED ISSUE #292: Adjust hpc-stack to adopt the latest system updates in WCOSS2 machine (Acron)
Recently, the lmod and wcoss2 system have been updated. The lmod file "startLmod" has been removed. The default PrgEnv-intel has been changed with the newly installed compiler, cray-pe has been renamed into craype. etc.
Current hpc-stack does not work on wcoss2.

Update and test with these new settings.

**Additional context**
A testing version has been created on WCOSS2.

Hang-Lei-NOAA:
This is for issue #292 
The new setting were used. 
libs/build_esmf.sh was updated with the correct WCOSS2 cray-mpich settings.


kgerheiser:
I would recommend adding this since we want FMS to build with AVX:

```
export STACK_fms_CFLAGS="-march=core-avx2"
export STACK_fms_FFLAGS="-march=core-avx2"
```

Hang-Lei-NOAA:
Okay, that is very necessary.

kgerheiser (approved):
Looks good.

libs/build_esmf.sh
```diff
@@ -113,7 +113,8 @@ case $MPI in
     export ESMF_COMM=${STACK_esmf_comm:-"mpich3"}
     ;;
   cray-mpich )
-    export ESMF_COMM=${STACK_esmf_comm:-"mpi"}
+    export ESMF_OS="Linux"
+    export ESMF_COMM=${STACK_esmf_comm:-"mpich3"}
```
aerorahul: I think this is going to break Gaea.
@climbfuji Can you please verity?

climbfuji: Yes, this will not work.

kgerheiser: What part will break it? `ESMF_COMM` or `ESMF_OS` (or both)?

Hang-Lei-NOAA: Keep the "export ESMF_COMM=${STACK_esmf_comm:-"mpi"}
Will break the wcoss2. (export in the conf does not work)
We have to finger out a way.

Hang-Lei-NOAA: Two system settings of cray-mpich are different.

kgerheiser: Setting the environment variable should work. 

Let's figure out why setting `STACK_esmf_comm` doesn't work because it should.

aerorahul: @Hang-Lei-NOAA 
Please add
`export STACK_esmf_comm="mpich3"` in `config_wcoss2.sh` 
This **should** work.

config/config_wcoss2.sh
```diff
@@ -51,8 +51,11 @@ export MPI_CXX=$SERIAL_CXX
 # Define the ESMF_COMM variable for WCOSS2
 # This is necessary to be done here rather than
 # stack_noaa.yaml, to keep one YAML file for NOAA.
-export STACK_esmf_comm="mpich3"
-export STACK_esmf_os="Linux"
+#export STACK_esmf_comm="mpich3"
```
aerorahul:
```suggestion
export STACK_esmf_comm="mpich3"
```

libs/build_esmf.sh
Hang-Lei-NOAA: I tryed that is not working. so I changed it.

Hang-Lei-NOAA: Kyle, could you please try it again?

libs/build_esmf.sh
```diff
@@ -113,7 +113,8 @@ case $MPI in
     export ESMF_COMM=${STACK_esmf_comm:-"mpich3"}
     ;;
   cray-mpich )
-    export ESMF_COMM=${STACK_esmf_comm:-"mpi"}
+    export ESMF_OS="Linux"
```
aerorahul: This one is tricky and should not be necessary.
If it is indeed necessary, please replace with:
```
export ESMF_OS=${STACK_esmf_os:-$host}
```

aerorahul (changes requested):
Some of these changes are going to break other machines.

libs/build_esmf.sh
climbfuji: The ESMF_OS line also needs to be removed, and possibly an option to overwrite the default setting "Unicos" needs to be added.

kgerheiser: I would like to, but I'm having a problem with my WCOSS account and I can't login. I'm waiting for a response from the help desk.

kgerheiser: Try printing out the contents of the variable in the build script and see if it's set.

libs/build_esmf.sh
Hang-Lei-NOAA: Good suggestion, the export ESMF_OS=${STACK_esmf_os:-$host} works

libs/build_esmf.sh
Hang-Lei-NOAA: After trying different combinations.
The issue is on ESMF_OS. It has to be there.
Taking Rahul's suggestion by using" export ESMF_OS=${STACK_esmf_os:-$host}" works

Hang-Lei-NOAA: Change default ESMF_COMM back to mpi for gaea, and export it to mpich3 in config file for wcoss2.

Hang-Lei-NOAA: @climbfuji Could you please try if adding "export ESMF_OS=${STACK_esmf_os:-$host}" will affact the gaea? Thanks

config/config_wcoss2.sh
```diff
@@ -52,7 +52,10 @@ export MPI_CXX=$SERIAL_CXX
 # This is necessary to be done here rather than
 # stack_noaa.yaml, to keep one YAML file for NOAA.
 export STACK_esmf_comm="mpich3"
-export STACK_esmf_os="Linux"
+#export STACK_esmf_os="Linux"
```
aerorahul: Please remove commented line.

config/config_wcoss2.sh
```diff
@@ -52,7 +52,10 @@ export MPI_CXX=$SERIAL_CXX
...
+#FMS to build with AVX:
+export STACK_fms_CFLAGS="-march=core-avx2"
+export STACK_fms_FFLAGS="-march=core-avx2"
 
 # LMod is coming to WCOSS2
```
aerorahul: If this is no longer the case, then please remove.

aerorahul (commented):
minor commnets.
otherwise looks good.

libs/build_esmf.sh
climbfuji: I am trying this now.

aerorahul (approved):
Looks good.
Waiting on @climbfuji to confirm this works on Gaea.

climbfuji:
> Looks good.
> Waiting on @climbfuji to confirm this works on Gaea.

Will take a little. I need to not only compile the code, but also compile and run the model, because some of the problems only show up when initializing ESMF in the model.

climbfuji:
> > Looks good.
> > Waiting on @climbfuji to confirm this works on Gaea.
> 
> Will take a little. I need to not only compile the code, but also compile and run the model, because some of the problems only show up when initializing ESMF in the model.

It doesn't even build:
```
ifort: command line warning #10418: option -reentrancy=threaded requires pthread library which cannot be statically linked.  Reverting to -reentrancy=none
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: cannot find -lmpi
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: cannot find -lmpi++
make[4]: *** [/lustre/f2/scratch/Dom.Heinzeller/hpc-stack/hpc-stack-wcoss2_update/pkg/ESMF_8_1_1/build/common.mk:2442: /lustre/f2/scratch/Dom.Heinzeller/hpc-stack/hpc-stack-wcoss2_update_install/intel-18.0.6.288/cray-mpich-7.7.11/esmf/8_1_1/bin/ESMF_PrintInfo] Error 1
make[4]: Leaving directory '/lustre/f2/scratch/esrl/Dom.Heinzeller/hpc-stack/hpc-stack-wcoss2_update/pkg/ESMF_8_1_1/src/apps/ESMF_PrintInfo'
make[3]: *** [/lustre/f2/scratch/Dom.Heinzeller/hpc-stack/hpc-stack-wcoss2_update/pkg/ESMF_8_1_1/build/common.mk:3661: tree] Error 1
make[3]: Leaving directory '/lustre/f2/scratch/esrl/Dom.Heinzeller/hpc-stack/hpc-stack-wcoss2_update/pkg/ESMF_8_1_1/src/apps'
make[2]: *** [/lustre/f2/scratch/Dom.Heinzeller/hpc-stack/hpc-stack-wcoss2_update/pkg/ESMF_8_1_1/build/common.mk:2425: build_apps] Error 2
make[2]: Leaving directory '/lustre/f2/scratch/esrl/Dom.Heinzeller/hpc-stack/hpc-stack-wcoss2_update/pkg/ESMF_8_1_1'
make[1]: *** [makefile:643: install_apps] Error 2
make[1]: Leaving directory '/lustre/f2/scratch/esrl/Dom.Heinzeller/hpc-stack/hpc-stack-wcoss2_update/pkg/ESMF_8_1_1'
make: *** [makefile:681: install] Error 2
BUILD FAIL!  Lib: esmf Error:2
```

climbfuji:
@Hang-Lei-NOAA @aerorahul gaea needs this:
```
ESMF_OS:                Unicos
```
I think your PR will work if I set
```
export STACK_esmf_os="Unicos"
```
in the gaea config, let me try.

aerorahul:
Ugh. 
@Hang-Lei-NOAA
Can you try by removing that line completely? 
i.e unset ESMF_OS and let ESMF auto detect.

Hang-Lei-NOAA:
After removing that line, no matter export it on config file or not.
It is not working on wcoss2.

kgerheiser:
How about conditionally setting it, if it's not empty?

climbfuji:
> How about conditionally setting it, if it's not empty?

If I make the following change, I can use Hang's branch as is:
```
diff --git a/config/config_gaea.sh b/config/config_gaea.sh
index 113b55d..c0dca57 100644
--- a/config/config_gaea.sh
+++ b/config/config_gaea.sh
@@ -31,5 +31,8 @@ export MPI_CC=cc
 export MPI_CXX=CC
 export MPI_FC=ftn

+export STACK_esmf_os="Unicos"
+export STACK_esmf_comm="mpi"
+
 # Load lmod environment
 source /lustre/f2/pdata/esrl/gsd/contrib/lua-5.1.4.9/init/init_lmod.sh
```

Note that only the first line is required, the second is just for completeness. It is inconsistent, however, because `libs/build_esmf.sh` now contains the correct default for gaea for `ESMF_COMM`, but not for `ESMF_OS`. I think it should be one way or the other. For example, the defaults in `libs/build_esmf.sh` could be `ESMF_OS=Unicos` and `ESMF_COMM=mpi`, and one would overwrite both for wcoss2 in the wcoss2 config. Or vice versa.

aerorahul:
This is from the ESMF build documentation. 
```
ESMF_OS
Possible value: output of uname -s except when cross-compiling or for UNICOS/mp where ESMF_OS is Unicos.
Not normally set by user unless cross-compiling. This variable indicates the target system for which the ESMF library is being built. Under normal circumstances, i.e. ESMF is being build on the target system, ESMF_OS is set automatically. However, when cross-compiling for a different target system ESMF_OS must be set to the respective target OS. For example, when compiling for the Cray X1 on an interactive X1 node ESMF_OS will be set automatically. However, when ESMF is being cross-compiled for the X1 on a Linux host the user must set ESMF_OS to Unicos manually in order to indicate the intended target platform.
```

Neither Gaea nor WCOSS2 are `normal` machines. Blame it on the Cray's. 

`build_esmf.sh` knows nothing about what host is being operated on.

aerorahul:
@Hang-Lei-NOAA 
What happens when you set it to Unicos for WCOSS2?

Hang-Lei-NOAA:
It is a message that mpich3 is not available

Hang-Lei-NOAA:
The combination of #export STACK_esmf_os="Linux" in config file 
and
set STACK_esmf_os="Unicos" in build_esmf 
works

climbfuji (approved):

PR #309, opened 2021-08-20
make the installation directory prefix and the modulefile readonly

### LINKED ISSUE #306: Make installed software read-only to discourage accidental rewrites?
It has been suggested that we make our installed hpc-stack read-only to prevent accidental overwrites.

Another, more sophisticated solution would be to use a unix group which had write permissions. hpc-stack updaters would have to specifically invoke the group name to get the permissions.

@arunchawla-NOAA this was one of the items mentioned at today's meeting...

aerorahul:
This PR makes the installation directory `prefix` and the modulefile for the package read only.
This will hopefully prevent accidental deletions/movements/additions and provide stability to the installation stack.

The owner will have to take drastic steps to deliberately remove an installation.

With great power comes great responsibility.

Fixes #306

### DIFF
--- a/libs/build_atlas.sh
+++ b/libs/build_atlas.sh
@@ -55,6 +55,9 @@ VERBOSE=$MAKE_VERBOSE make -j${NTHREADS:-4}
 [[ $MAKE_CHECK =~ [yYtT] ]] && ctest
 VERBOSE=$MAKE_VERBOSE $SUDO make install
 
+# Make the installation prefix read-only for all
+$MODULES && $SUDO chmod a-w $prefix
+
 # generate modulefile from template
 $MODULES && update_modules mpi $name $repo-$version
 echo $name $version $URL >> ${HPC_STACK_ROOT}/hpc-stack-contents.log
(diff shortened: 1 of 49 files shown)


kgerheiser:
Should `OVERWRITE` override this? Or another option? I spin up and delete a lot of hpc-stack builds in my development environment.

Should `chmod` run recursively `-R`?

aerorahul:
> Should `OVERWRITE` override this? Or another option? I spin up and delete a lot of hpc-stack builds in my development environment.
> 
> Should `chmod` run recursively `-R`?

`OVERWRITE` is in direct conflict with this now, because this is essentially preventing an overwrite.

To restore the `OVERWRITE` option, we will have to first do `chmod u+w $prefix` from the outside and then run the stack script that will allow removing the `$prefix`

This is going to be super annoying for power users.

Ideally, this is unnecessary.   It being put in place because of manual moving stuff around that should not be touched.
So everyone is going to be punished because of this.

I don't think we need to go recursively down the installation prefix, yet anyways.

aerorahul:
will reopen later after resolving conflicts.

PR #318, opened 2021-09-03
Download and install CRTM fix files.

aerorahul:
This PR:
- downloads and installs CRTM fix files alongside the library
- sets an environment variable `CRTM_FIX` when the `crtm` module is loaded.

### DIFF
--- a/libs/build_nceplibs.sh
+++ b/libs/build_nceplibs.sh
@@ -215,15 +215,17 @@ if [[ "$name" == "crtm" ]]; then
   if [[ ${STACK_crtm_install_fix:-} =~ [yYtT] ]]; then
     if [[ ! -d crtm_fix-$version ]]; then
       crtm_tarball=fix_REL-${install_as}_emc.tgz
+      rm -f $crtm_tarball
       $WGET ftp://ftp.ucar.edu/pub/cpaess/bjohns/$crtm_tarball
       tar xzf $crtm_tarball
       mv fix crtm_fix-$version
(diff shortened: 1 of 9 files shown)


libs/build_nceplibs.sh
```diff
@@ -223,6 +246,17 @@ VERBOSE=$MAKE_VERBOSE make -j${NTHREADS:-4}
 [[ $USE_SUDO =~ [yYtT] ]] && sudo -- bash -c "export PATH=$PATH; make install" \
                           || make install
 
+cd ${HPC_STACK_ROOT}/${PKGDIR:-"pkg"}
+
+# Install CRTM fix files
+if [[ "$name" == "crtm" ]]; then
+  if [[ ${STACK_crtm_install_fix:-} =~ [yYtT] ]]; then
+    if [[ -d crtm_fix-$version ]]; then
+      ./link_crtm_coeffs.sh ./crtm_fix-$version $prefix/fix
```
kgerheiser, line 255: Does there need to be a check here if it's already installed?

aerorahul: how will it be already installed? `$prefix` is the location of the library, which is protected by `OVERWRITE`.
This is the problem remember, the coefficients living with the library.

kgerheiser: Right, the top-level overwrite is there.

kgerheiser (approved):

edwardhartnett:
Thanks this is great!

Issue #319, opened 2021-09-07
Add MAPL under ESMF in module hierarchy


kgerheiser:
MAPL should be under ESMF in the module hierarchy to ensure that the builds are compatible. UFS ran into an issue where they were using ESMF 8.2.0 beta 14 while MAPL was built with ESMF 8.1.1 and it caused errors due to a change in the ESMF API.

kgerheiser:
Fix with #325

PR #325, opened 2021-09-09
Add install_as option to MAPL

kgerheiser:
Related #319 

My naive way of adding a hierarchy didn't work well. Add an install_as option to MAPL so modules can at least be named.

### DIFF
--- a/libs/build_mapl.sh
+++ b/libs/build_mapl.sh
@@ -76,6 +76,6 @@ VERBOSE=$MAKE_VERBOSE make -j${NTHREADS:-4} install
 # generate modulefile from template
 modpath=mpi
 
-module_substitutions="-DMAPL_ESMF_VERSION=${short_esmf_ver}"
+module_substitutions="-DMAPL_ESMF_VERSION=${ESMF_VERSION:-}"
 $MODULES && update_modules $modpath $name $install_as "" $module_substitutions
 echo $name $id $URL >> ${HPC_STACK_ROOT}/hpc-stack-contents.log
--- a/libs/build_mapl.sh
+++ b/libs/build_mapl.sh
@@ -9,7 +9,7 @@ version=${2:-${STACK_mapl_version:-"main"}}
 # Hyphenated version used for install prefix
 compiler=$(echo $HPC_COMPILER | sed 's/\//-/g')
 mpi=$(echo $HPC_MPI | sed 's/\//-/g')
-id=${version//\//-}
+id=$(echo $version | sed 's/v//')
 
 if $MODULES; then
   set +x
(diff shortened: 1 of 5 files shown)


climbfuji (approved):
This looks good to me, thanks for adding the `install_as` option for MAPL.

aerorahul (approved):

kgerheiser:
Made some changes. 

Added a `esmf_version` option to MAPL, and an `always_load(ESMF/<version>)` to the MAPL module.

kgerheiser:
Hmm, that won't work with a debug build of ESMF. Might have to forego the module version check.

kgerheiser:
And let's autogenerate the MAPL module with the ESMF version used to build it rather than manually setting it with `install_as` (which is still possible).

stack_helpers.sh
```diff
@@ -58,6 +58,7 @@ EOF
     CMAKE_OPTS="-DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}"
     CMAKE_OPTS+=" -DTMPL_FILE=$tmpl_file -DVERSION=$version"
     [[ -n "${py_version:-}" ]] && CMAKE_OPTS+=" -DPYTHON_VERSION=$py_version"
+    CMAKE_OPTS+=" -DMAPL_ESMF_VERSION=${ESMF_VERSION:-}"
```
aerorahul: There is nothing in `stack_helpers.sh` specific to any particular library.
Can this be kept out of `stack_helpers.sh`

aerorahul: I have to protest for this hack.

aerorahul (changes requested):
`update_modules.sh` in `stack_helpers.sh` is generic for all modules and this change is too specific a hack to fit in a generic function.

libs/build_mapl.sh
```diff
@@ -73,5 +75,5 @@ VERBOSE=$MAKE_VERBOSE make -j${NTHREADS:-4} install
 
 # generate modulefile from template
 modpath=mpi
-$MODULES && update_modules $modpath $name $id
+$MODULES && update_modules $modpath $name $install_as "" " -DMAPL_ESMF_VERSION=${ESMF_VERSION}"
```
aerorahul: I like this approach.
1 suggestion:
```suggestion
$MODULES && update_modules $modpath $name $install_as "" "-DMAPL_ESMF_VERSION=${ESMF_VERSION}"
```

stack_helpers.sh
```diff
@@ -58,6 +59,7 @@ EOF
     CMAKE_OPTS="-DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}"
     CMAKE_OPTS+=" -DTMPL_FILE=$tmpl_file -DVERSION=$version"
     [[ -n "${py_version:-}" ]] && CMAKE_OPTS+=" -DPYTHON_VERSION=$py_version"
+    CMAKE_OPTS+="${module_substitutions}"
```
aerorahul: Going with the suggestion from `build_mapl.sh`, as this will allow the user to be a bit lax on having to enforce a ` ` in the build script.
```suggestion
    CMAKE_OPTS+=" ${module_substitutions}"
```

aerorahul (commented):
This looks better.

aerorahul (approved):

PR #326, opened 2021-09-09
bugfix for hera miniconda installation.

aerorahul:
Hera admins provide a mirror to the anaconda repo, and users have to use that to install miniconda.
This PR:
- allows setting the url for hera while reverting to anaconda for all others.
- fixes a bug for multiuser installation of miniconda.
- versions of packages are an issue on Hera because that mirror is selective and to get all possible options on the Hera miniconda mirror will be a sysadmin requirement and will probably never be done in any reasonable time on Hera.  Best to use the available one.
- added hofxdiag.yaml conda/pyvenv requirement file.  This adds cartopy, which will need another PR for downloading cartopy shapefiles and pointing to it. 

This PR also:
- disables CI for every push.  It retains for PR
- adds a on-demand option to run CI with `workflow_dispatch` option.


kgerheiser:
It all looks fine to me though I've never really used the Python stuff.

I like the change in the Actions. I spent a bunch of time clicking to cancel workflows today.

kgerheiser (approved):

Issue #332, opened 2021-09-21
wgrib2 module missing in MPI hierarchy for CMake build


kgerheiser:
**Describe the bug**
wgrib2 should be able to be built with CMake using `is_nceplib`, but the module was removed from the MPI hierarchy 

**To Reproduce**
Attempt to use CMake build of wgrib2 such as in stack_ufs_weather_ci.yaml with modules

**Expected behavior**
Module should be installed

**System:**
All

aerorahul:
I think we have decided to not maintain a cmake build of wgrib2. 
This ticket is obsolete. 
Closing.

PR #333, opened 2021-09-21
Nceplibs updates

### LINKED ISSUE #314: EMC_Post repo being renamed to UPP
EMC_Post is being renamed to UPP on Github next week (Sept 6)

[According to Github](https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/renaming-a-repository) this will have no impact on hpc-stack as the current remote will continue to work, but it should be changed when the switch is made.

This line will need to be changed: https://github.com/NOAA-EMC/hpc-stack/blob/fe6fa12a29fb33adf8c454968d510ac718972776/libs/build_nceplibs.sh#L185

kgerheiser:
* Update g2 and grib_util for bugfixes from WCOSS2 transition
* Add wgrib2 module to MPI hierarchy to support CMake build
* Update ufs_weather_ci to match what ufs-weather-model uses in develop
* Update to new UPP URL and version to 10.0.9

Fix #314 and #332

### DIFF
--- a/modulefiles/mpi/compilerName/compilerVersion/mpiName/mpiVersion/wgrib2/wgrib2.lua
+++ b/modulefiles/mpi/compilerName/compilerVersion/mpiName/mpiVersion/wgrib2/wgrib2.lua
@@ -5,21 +5,27 @@ local pkgName = myModuleName()
 local pkgVersion = myModuleVersion()
 local pkgNameVer = myModuleFullName()
 
-local hierA        = hierarchyA(pkgNameVer,1)
-local compNameVer  = hierA[1]
+local hierA        = hierarchyA(pkgNameVer,2)
+local mpiNameVer   = hierA[1]
+local compNameVer  = hierA[2]
+local mpiNameVerD  = mpiNameVer:gsub("/","-")
 local compNameVerD = compNameVer:gsub("/","-")
 
 conflict(pkgName)
 
+load("netcdf")
+prereq("netcdf")
+
 local opt = os.getenv("HPC_OPT") or os.getenv("OPT") or "/opt/modules"
 
-local base = pathJoin(opt,compNameVerD,pkgName,pkgVersion)
+local base = pathJoin(opt,compNameVerD,mpiNameVerD,pkgName,pkgVersion)
 
 prepend_path("PATH", pathJoin(base,"bin"))
 setenv("wgrib2_ROOT", base)
 setenv("wgrib2_VERSION", pkgVersion)
 setenv("WGRIB2_INC", pathJoin(base,"include"))
-setenv("WGRIB2_LIB", pathJoin(base,"lib","libwgrib2.a"))
+setenv("WGRIB_LIB", pathJoin(base,"lib/libwgrib2.a"))
+setenv("WGRIB2_LIBAPI", pathJoin(base,"lib/libwgrib2_api.a"))
 
 whatis("Name: ".. pkgName)
 whatis("Version: " .. pkgVersion)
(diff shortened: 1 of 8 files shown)


modulefiles/mpi/compilerName/compilerVersion/mpiName/mpiVersion/wgrib2/wgrib2.lua
```diff
@@ -0,0 +1,27 @@
+help([[
+]])
+
+local pkgName = myModuleName()
+local pkgVersion = myModuleVersion()
+local pkgNameVer = myModuleFullName()
+
+local hierA        = hierarchyA(pkgNameVer,1)
+local compNameVer  = hierA[1]
+local compNameVerD = compNameVer:gsub("/","-")
```
aerorahul, line 12: This is a module file in the MPI hierarchy, but the paths here are for the compiler hierarchy.

modulefiles/mpi/compilerName/compilerVersion/mpiName/mpiVersion/wgrib2/wgrib2.lua
```diff
@@ -0,0 +1,27 @@
...
+
+conflict(pkgName)
+
+local opt = os.getenv("HPC_OPT") or os.getenv("OPT") or "/opt/modules"
+
+local base = pathJoin(opt,compNameVerD,pkgName,pkgVersion)
```
aerorahul: I don't think this is consistent with the location of the modulefile

aerorahul (changes requested):
I think there is a mismatch in the wgrib2 modulefile.

Do we need to build wgrib2 in the MPI hierarchy at all?  I am not sure, but then I am also not a user of this.

aerorahul:
@kgerheiser Sorry for the delay in reviewing this.

modulefiles/mpi/compilerName/compilerVersion/mpiName/mpiVersion/wgrib2/wgrib2.lua
kgerheiser: For the CMake build of wgrib2 in `build_nceplibs`, which uses external NetCDF it was under MPI. If you want to build that it will fail to generate the module.

aerorahul: Ok. Then, shouldn't these lines (8-10) be: 
```
local hierA        = hierarchyA(pkgNameVer,2)
local mpiNameVer   = hierA[1]
local compNameVer  = hierA[2]
local mpiNameVerD  = mpiNameVer:gsub("/","-")
local compNameVerD = compNameVer:gsub("/","-")
```

modulefiles/mpi/compilerName/compilerVersion/mpiName/mpiVersion/wgrib2/wgrib2.lua
aerorahul: and this line should then be:
```
local base = pathJoin(opt,compNameVerD,mpiNameVerD,pkgName,pkgVersion)
```

aerorahul (changes requested):
See other MPI hierarchy modulefiles as an example for wgrib2 under mpi tree.

aerorahul (commented):
how do you distinguish the `cmake` build of wgrib2 from the non-cmake build.
I see we have `build_wgrib2.cmake` which builds only the serial version.

Do we even need to support the `cmake` build and therefore the build of wgrib2 in `build_nceplibs.sh`?

aerorahul (approved):
Looks good.

PR #344, opened 2021-10-06
Add support for serial w3emc (v2.9.0)

kgerheiser:
Need a serial version of w3emc for WCOSS2. Starting with version 2.9.0 w3emc has no dependencies.


modulefiles/compiler/compilerName/compilerVersion/w3emc/w3emc.lua
```diff
@@ -0,0 +1,30 @@
...
+
+local hierA        = hierarchyA(pkgNameVer,1)
+local compNameVer  = hierA[1]
+local compNameVerD = compNameVer:gsub("/","-")
+
+conflict(pkgName)
+
+local opt = os.getenv("HPC_OPT") or os.getenv("OPT") or "/opt/modules"
+
+local base = pathJoin(opt,compNameVerD,mpiNameVerD,pkgName,pkgVersion)
```
aerorahul:
```suggestion
local base = pathJoin(opt,compNameVerD,pkgName,pkgVersion)
```

libs/build_nceplibs.sh
```diff
@@ -30,7 +30,7 @@ if $MODULES; then
 
   case $name in
     # The following require MPI
-    nemsio | nemsiogfs | ncio | nceppost | upp | w3emc)
+    nemsio | nemsiogfs | ncio | nceppost | upp)
```
aerorahul, line 33:
```suggestion
   # Starting with version 2.9.0 w3emc no longer relies on MPI
   w3emc)
      version_number=$(echo $version | cut -c 2-)
      major_ver=$(echo $version_number | cut -d. -f1)
      minor_ver=$(echo $version_number | cut -d. -f2)
      using_mpi=NO
      if [[ "$major_ver" -le "2" ]]; then
          if [[ "$minor_ver" -lt "9" ]]; then
              module load hpc-$HPC_MPI
              using_mpi=YES
          fi
      fi
      ;;
    nemsio | nemsiogfs | ncio | nceppost | upp)
```

libs/build_nceplibs.sh
```diff
@@ -79,9 +79,23 @@ if $MODULES; then
...
+              module load hpc-$HPC_MPI
+              module load netcdf
+              module load sigio
+              module load nemsio
+          else
+              using_mpi=NO
+          fi
+      else
+          using_mpi=NO
+      fi
```
aerorahul:
```suggestion
      # Only MPI version of w3emc requires these: 
      if [[ "$using_mpi" =~ [yYtT] ]]; then
              module load netcdf
              module load sigio
              module load nemsio
      fi
```

aerorahul (changes requested):
A couple of separations to align with how the rest of the script is laid out, and more importantly, 1 bug caught.

aerorahul (approved):
looks good.

Hang-Lei-NOAA (approved):

PR #346, opened 2021-10-07
Use CMAKE_INSTALL_LIBDIR for serial w3emc

### LINKED ISSUE #345: Serial w3emc module needs to use CMAKE_INSTALL_LIBDIR
**Describe the bug**
The w3emc module hardcodes "lib" when w3emc uses CMAKE_INSTALL_LIBDIR which could be lib64

**To Reproduce**
Build w3emc 2.9.1 with modules

**Expected behavior**
Module should point to correct libdir

**System:**
Acorn

kgerheiser:
Fix #345


aerorahul (approved):

Hang-Lei-NOAA (approved):

edwardhartnett:
It's not clear to me why a serial build of w3emc should be needed. Since it does not include any MPI calls, it should be the same as an MPI build of the library...

kgerheiser:
@edwardhartnett this is compared to version v2.7.3 of w3emc where it still had gblevents using Nemsio and MPI, and therefore the module was under MPI.

Hang-Lei-NOAA:
It was reported to be used in a MPMD run, which requires serial code.

PR #351, opened 2021-10-18
Install CRTM fix files with cp instead of mv

### LINKED ISSUE #349: HPC_MPI unset when building nccmp
**Describe the bug**
With this PR that builds serial netcdf (https://github.com/NOAA-EMC/hpc-stack/commit/bdecd1b15bd5a320cbc5dfd0b181e0264fad16b4), HPC_MPI is unset when building the serial builds, but then that variable is used in nccmp, cdo, and nco.

See: https://github.com/NOAA-EMC/hpc-stack/blob/b501323ec3030a2721350e87d41f708f8441a462/libs/build_nccmp.sh#L18

Instead, this:

https://github.com/NOAA-EMC/hpc-stack/blob/b501323ec3030a2721350e87d41f708f8441a462/build_stack.sh#L151

Should be set to an empty string:

`export HPC_MPI=""`

### LINKED ISSUE #350: CRTM fix installed with mv instead of cp causing subsequent installs to fail
**Describe the bug**
When installing CRTM fix files the files are moved instead of copied. So, if you attempt to install CRTM again the fix files in the `pkg` dir aren't there anymore.

Instead of `mv`, the files should be copied with `cp`:

https://github.com/NOAA-EMC/hpc-stack/blob/b501323ec3030a2721350e87d41f708f8441a462/libs/build_nceplibs.sh#L241

**To Reproduce**

Install fix files twice with CRTM

kgerheiser:
Fix #350 

Fix #349

### DIFF
--- a/build_stack.sh
+++ b/build_stack.sh
@@ -148,7 +148,7 @@ if $MODULES; then
 
   # Save $HPC_MPI variable
   _HPC_MPI=$HPC_MPI
-  unset HPC_MPI
+  export HPC_MPI=""
 
   # Build hdf5 and netcdf as serial versions
   build_lib hdf5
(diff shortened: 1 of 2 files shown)


aerorahul (approved):
LGTM

PR #352, opened 2021-10-22
Load bacio when building w3emc

kgerheiser:
With bacio 2.9.2 w3emc depends on bacio

### DIFF
--- a/libs/build_nceplibs.sh
+++ b/libs/build_nceplibs.sh
@@ -91,6 +91,7 @@ if $MODULES; then
       module load nemsio
       ;;
     w3emc)
+      module load bacio
       if [[ "$using_mpi" =~ [yYtT] ]]; then
           module load netcdf
           module load sigio


aerorahul (approved):

aerorahul:
This is approved, but the yamls are not installing w3emc 2.9.2 in this branch.
There is a bit of round robin going on wrt w3nco, w3emc and nemsio.
This is going to make a lot of mess and quickly throw exceptions if this isn't resolved.

edwardhartnett:
What do we need to do to resolve it?

aerorahul:
@edwardhartnett See #353 
There are a lot of `if this version do this or do that` in `build_nceplibs.sh` that is going on as a result of these recent library upgrades.  The library ugrades are necessary and we should encourage people to move to the latest ones as we release them.

For e.g.
nemsio used to depend on w3nco.
we merged w3nco into w3emc and deprecated w3nco.
nemsio now depends on w3emc.

w3emc used to have a subroutine `glbevents` that is now pulled out.  This routine depends on nemsio. 
So we had to move the build of nemsio after w3emc for older versions of nemsio, but w3emc needs to be built before nemsio for newer versions of nemsio.

It is confusing.

Meanwhile, WCOSS2 libraries are a hodge-podge of new and old.

My 2c of recommendation would be to move all libraries to the newest possible versions. But I know NCO will likely object for accepted or transitioned applications.  Note, they have been accepted on a machine that has not been itself accepted. Meanwhile, there are still a lot of applications (major applications) that have not transitioned over.

edwardhartnett:
As you note we are going through some necessary but annoying library changes. There are some libraries being combined, glbevents has been moved, and the bacio library has changed names. And there are more changes to come.

We cannot demand that everyone upgrade, unfortunately. We need to make each package cope with the older versions if possible. If not possible, we need the build systems to be smart enough to know that they don't have what they need.

(Another area of confusion is that CMake does not do a great job when multiple libraries depend on the same thing, and then on each other. For example, g2 needs bacio, UFS_UTILS needs both bacio and g2. If I give UFS_UTILS a newer version of bacio than g2 got, UFS_UTILS will downgrade to the version used to build g2. I'm not sure if this is the correct thing for it to do, but I don't know what else it should do instead. Anyway, this is another area of confusion.)

I think that after some more testing time we should upgrade hpc-stack to the new versions of all the NCEPLIBS. I think that's what we typically do - update hpc-stack when a new NCEPLIBS upgrade happens.

aerorahul:
The CMake "issue" notes is not an cmake issue. It is how cmake is used in our systems. 

We do a blanket find package or find dependency disregarding if the package was already found.

aerorahul:
I also don't understand how it will downgrade, if bacio_ROOT is uniquely set upfront before cmake is called.

kgerheiser:
This PR should be fine. It will just require bacio to be loaded before w3emc, even if w3emc doesn't always use it (i.e. w3emc < 2.9.2)

kgerheiser:
I also don't really understand this.

> If I give UFS_UTILS a newer version of bacio than g2 got, UFS_UTILS will downgrade to the version used to build g2. I'm not sure if this is the correct thing for it to do, but I don't know what else it should do instead. Anyway, this is another area of confusion.)

It runs `find_dependency` and they will both find the same version of bacio. Once g2 is built is has no memory of what version of bacio it was built with, only what's in your environment at the moment. 

But yeah, it will find bacio multiple times. We could add a check in the package configs that if the target already exists it won't find it twice. However, this behavior is more of a quirk and doesn't really have any actual effect.

edwardhartnett:
It seems to find the old version of bacio in the g2 config file. That is, it loads the config file for g2, and that sets the flags which point to an older version of bacio.

What is the bacio_ROOT environment variable you speak of? Must we manually set environment variables for our build system? That's certainly not very groovy. Anyway, I am not doing that.

The build works for me if I am careful to use the same version of bacio for g2 and ufs_utils.

aerorahul:
One does not have to set that variable ofcourse.
The build system needs to know where to find the dependencies.
CMake looks in the usual places and places defined under `CMAKE_PREFIX_PATH`.
So if the dependencies are in some special place, CMake needs to know.
There are ways to tell CMake; add the Package installation path to `CMAKE_PREFIX_PATH` or `Package_ROOT` or provide `HINTS`.
If one doesn't do either, and the package to be found is in some unusual location e.g. `$HOME/Downloads`, CMake cannot divine that information.

How do "you" set a version for use without manually changing CMakeLists.txt in both repositories?

edwardhartnett:
I do this:

`cmake -DCMAKE_BUILD_TYPE="Debug" -DCMAKE_PREFIX_PATH="/usr/local/jasper-2.0.33" -DENABLE_DOCS=ON .. && make && make test`

I install every package into its own directory with a version number, so I can use any combination...

kgerheiser:
Would be interesting to see the output from CMake when you get two different version, like when it says `Found bacio: <path> version: x.y.z`

PR #364, opened 2021-11-17
Skip build instead of exiting with failure if library exists

### LINKED ISSUE #335: Overwrite Y/N logic should be redesigned to ignore existing installations

I'm always frustrated when ... I spend an entire day setting up the correct libraries for a given ufs-weather-model hash on a new system, because I can
(a) either go through each line in the yaml file manually and check what needs to be changed
(b) or install the default stack (`stack_noaa.yaml`) and then any missing items

For (b), I currently need to copy `stack_noaa.yaml` to a new yaml file, delete everything except the one library that I want to install with a different version number, then try to compile again, rinse and repeat. Or set `OVERWRITEY` and spend hours and hours reinstalling everything except the one line that I changed.

It would be much easier if the existing `OVERWRITE=Y|N` could be expanded so that an existing installation wouldn't cause `build_stack.sh` to abort, but instead ignore it and move on to the next item.

kgerheiser:
Use exit 0, instead of exit 1 if build already exists and overwrite is not enabled.

Fix #335

### DIFF
--- a/libs/build_atlas.sh
+++ b/libs/build_atlas.sh
@@ -27,8 +27,13 @@ if $MODULES; then
 
   prefix="${PREFIX:-"/opt/modules"}/$compiler/$mpi/$name/$repo-$version"
   if [[ -d $prefix ]]; then
-    [[ $OVERWRITE =~ [yYtT] ]] && ( echo "WARNING: $prefix EXISTS: OVERWRITING!"; $SUDO rm -rf $prefix ) \
-                               || ( echo "WARNING: $prefix EXISTS, SKIPPING"; exit 0 )
+      if [[ $OVERWRITE =~ [yYtT] ]]; then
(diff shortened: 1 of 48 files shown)


aerorahul (approved):

climbfuji (approved):
That's awesome, thanks for making these changes.

kgerheiser:
The parentheses need to be removed because `exit` will only exit out of that subshell and then continue building

aerorahul:
> The parentheses need to be removed because `exit` will only exit out of that subshell and then continue building

I don't know if the paranthese

> The parentheses need to be removed because `exit` will only exit out of that subshell and then continue building

I think we will have to rewrite the logic into the traditional `if-then-else` form instead of the compact use as is the case currently.

kgerheiser:
Yes, just tested that out. I'll use `if, else`

Issue #373, opened 2022-01-11
Add crtm v2.4.0 changes into hpc-stack and install it on all platforms


Hang-Lei-NOAA:
**Please describe the package or library you would like to add to hpc-stack.**
The crtm v2.4.0 EMC version has been released and tested in the past several weeks with Haixia, Ben and Hang. It has been added on hera installation of hpc-stack. Now the testing result on hera has confirmed with the expected. The major changes are on the fix files. The existing hpc-stack script has no problem on code build. But the source URL and script to processing the fix files need to be updated according to the changes. Install crtm v2.4.0 on all platforms are requested by Haixia.

**What applications at NOAA will be using this package or library?**
GSI, GDAS

**Is there already a package or library in hpc-stack that provides this, or related, functionality?**
Yes, but need to update the scripts to process fix files.

**Additional context**
Add any other context or screenshots about the package or library here.

**Will This Package be Needed in an Operational Application?**
Yes

**WCOSS System Software Request Checklist**

General questions:

Supervisor or sponsor of the requester

Software name and version, specific URL to the software

Software type - New or Upgrade existing?

Justification ( List NPS models using the software)

Completion Time requested

Software License including all Dependency Software Licenses 

Support contact(s) who must have a WCOSS account

Dependency Software list

Installation instructions

Test and verification instructions

Technical Review Checklist for open source software - Review the source code to answer the following questions

Licenses for the requested software and its dependencies

Licenses -  confirm the software Licenses are acceptable

Maturity

Acceptable - Stable, production, or equivalent

Self-contained

No external http, https,  ftp, or other URI exists except that in comments

No binary files in the package unless they are in the approved list

No publicly disclosed cybersecurity vulnerabilities and exposures 

Searching https://cve.mitre.org/cve/

Security High Level Checklists 

Is it prohibited by DHS/DOC/NOAA/NWS

Is it provided by a trusted source? Trusted sources include other NWS, NOAA, or DOC, agencies, or other Federal agencies that operate at a FISMA high or equivalent level. Additionally, trusted sources could be third-party agencies through which there is an existing SLA on file (such as RedHat). 

Is software support offered (is it being updated and patched). 

If maintained by a private entity, does the entity operate in a foreign country, especially a prohibited foreign country (China, Russia, Iran, North Korea, etc.). 

Is there sufficient documentation to support maintenance 

Are there known vulnerabilities or weaknesses

Is there a need for privileged processes 

[...]

PR #375, opened 2022-01-13
part of issue #373

Hang-Lei-NOAA:
Upon Ben's request, we access crtm code from the new fork repo in github/noaa-emc
https://github.com/NOAA-EMC/crtm
It is a fork of the github.com/JCSDA/crtm repository. That only include the EMC version for GSI use.

### DIFF
--- a/libs/build_nceplibs.sh
+++ b/libs/build_nceplibs.sh
@@ -94,6 +94,10 @@ if $MODULES; then
       module load sp
       module load ip2
       ;;
+    crtm)
+      module load hpc-$HPC_MPI
+      module load netcdf
+      ;;
     ip2)
       module load sp
       ;;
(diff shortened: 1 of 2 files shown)


arunchawla-NOAA:
So probably a silly set of questions but ...
1. Why dont we just use the JCSDA repository?
2. Who is managing the EMC fork ?

Hang-Lei-NOAA:
Ben (Benjamin johnson) is managing this repo under NOAA-EMC.
He is the project manager.
I would think using the JCSDA repo is better, although that repo has too
many branches.

Hang-Lei-NOAA:
In previous emails, Ben said that " that'll be the easiest way to keep up
to date with latest releases."
I have asked him right now directly by changing to the JCSDA repo. Will see
what he will reply.

Hang-Lei-NOAA:
Ben's opinion on whether to use JCSDA or fork is up to EMC.
He pointed out: The JCSDA/crtm repository gets updated with each public
release (i.e., when the version number changes or when there's a
center-specific release), so I recommend a simple fork which will permit
EMC-specific modifications but while also being able to keep up to date
with new releases. It's also a good place to create issues etc. that we're
sure to see.
So, we can make a decision on this. Using fork or JCSDA repo are the same
in the script, which I have tested. We only need to change the URL.

kgerheiser (approved):

PR #381, opened 2022-02-09
feature/linux centos7 intel

natalie-perlin:
Updates to build the hpc-stack on RedLine cluster based on Linux CenOS7, intel compilers and impi, modules: ics/2020.0 , mpi/impi/2020u0
(Natalie Perlin nperlin@redlineperf.com)

### DIFF
--- a/libs/build_nceplibs.sh
+++ b/libs/build_nceplibs.sh
@@ -173,6 +173,7 @@ if $MODULES; then
       fi
       ;;
   esac
+  module try-load cmake
   module list
   set -x
 
@@ -232,7 +233,7 @@ case $name in
     extraCMakeFlags="-DBUILD_POSTEXEC=OFF"
     ;;
   crtm)
-    URL="https://github.com/noaa-emc/emc_crtm"
+    URL="https://github.com/NOAA-EMC/crtm.git"
     ;;
   wgrib2)
     extraCMakeFlags="${STACK_wgrib2_cmake_opts:-}"
@@ -262,7 +263,9 @@ software=$name-$version
 if [[ ! -d $software ]]; then
   git clone $URL $software
   cd $software
-
+  if [[ "$name" == "crtm" ]]; then
+    version=release/REL-${install_as}_emc
+  fi
   git checkout $version
   git submodule update --init --recursive
 fi
@@ -272,19 +275,21 @@ cd ${HPC_STACK_ROOT}/${PKGDIR:-"pkg"}
 # Download CRTM fix files
 if [[ "$name" == "crtm" ]]; then
   if [[ ${STACK_crtm_install_fix:-} =~ [yYtT] ]]; then
-    if [[ ! -d crtm_fix-$version ]]; then
+    if [[ ! -d crtm_fix-${install_as} ]]; then
       crtm_tarball=fix_REL-${install_as}_emc.tgz
       rm -f $crtm_tarball
       $WGET ftp://ftp.ucar.edu/pub/cpaess/bjohns/$crtm_tarball
       tar xzf $crtm_tarball
-      mv fix crtm_fix-$version
+      mv fix crtm_fix-${install_as}
       rm -f $crtm_tarball
     fi
-    if [[ ! -f link_crtm_coeffs.sh ]]; then
-      $WGET https://raw.githubusercontent.com/NOAA-EMC/GSI/master/ush/link_crtm_coeffs.sh
-      sed -i'.backup' -e 's/LINK="ln -sf"/LINK="cp"/g' link_crtm_coeffs.sh
-      chmod +x link_crtm_coeffs.sh
-      rm -f link_crtm_coeffs.sh.backup
+    if [[ "${install_as}" == "2.3.0" ]]; then
+     if [[ ! -f link_crtm_coeffs.sh ]]; then
+       $WGET https://raw.githubusercontent.com/NOAA-EMC/GSI/master/ush/link_crtm_coeffs.sh
+       sed -i'.backup' -e 's/LINK="ln -sf"/LINK="cp"/g' link_crtm_coeffs.sh
+       chmod +x link_crtm_coeffs.sh
+       rm -f link_crtm_coeffs.sh.backup
+     fi
     fi
   fi
 fi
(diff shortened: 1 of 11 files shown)


kgerheiser:
I don't think your branch is up-to-date with the CRTM changes from a week or so ago

libs/build_nceplibs.sh
```diff
@@ -232,7 +232,7 @@ case $name in
     extraCMakeFlags="-DBUILD_POSTEXEC=OFF"
     ;;
   crtm)
-    URL="https://github.com/NOAA-EMC/crtm.git"
+    URL="https://github.com/noaa-emc/emc_crtm"
```
kgerheiser: Are these changes from an outdated branch?

libs/build_proj.sh
```diff
@@ -41,7 +41,7 @@ cd ${HPC_STACK_ROOT}/${PKGDIR:-"pkg"}
 software=$name-$version
 URL="https://download.osgeo.org/proj/$software.tar.gz"
 
-[[ -d $software ]] || ( $WGET $URL; tar -xf $software.tar.gz )
+[[ -d $software ]] || ( $WGET --no-check-certificate $URL; tar -xf $software.tar.gz )
```
kgerheiser, line 45: Curious about what this certificate check is for?

libs/build_tiff.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
kgerheiser, line 1: What is tiff used for?

kgerheiser (changes requested):

libs/build_proj.sh
natalie-perlin: Without the --no-check-certificate several libraries are not allowed to download. The error message (similar for all the build_<library>.sh modified):

 wget -nv https://www.sqlite.org/2020/sqlite-autoconf-3320300.tar.gz
ERROR: cannot verify www.sqlite.org's certificate, issued by ‘/C=US/O=Let's Encrypt/CN=R3’:
  Issued certificate has expired.
To connect to www.sqlite.org insecurely, use `--no-check-certificate'.
BUILD FAIL!  Lib: sqlite Error:1

libs/build_tiff.sh
natalie-perlin: The tiff library is required for building  "proj" library. Modified the order of libraries requested to be built in build_stack.sh to reflect this dependency, such that:
...
build_lib sqlite
build_lib tiff
build_lib proj
...

libs/build_proj.sh
kgerheiser: Is that on Orion? I put in a ticket about that a few months ago.

natalie-perlin: This is on a Linux cluster of the RedLine Performance

libs/build_proj.sh
```diff
@@ -51,6 +52,14 @@ CMAKE_OPTS=${STACK_proj_cmake_opts:-""}
 
 [[ $MAKE_CHECK =~ [yYtT] ]] || CMAKE_OPTS+=" -DBUILD_TESTING=OFF"
 
+if [[ ! -z ${TIFF_ROOT} ]] ; then
+  export TIFF_LIBRARY=${TIFF_ROOT}/lib64/libtiff.so
```
kgerheiser: Do you know if this is always lib64? I think the tiff build is using `CMAKE_INSTALL_LIBDIR` which is `lib` or `lib64` depending on OS. So, I think this might need to be more generic.

libs/build_sqlite.sh
```diff
@@ -36,7 +36,7 @@ cd ${HPC_STACK_ROOT}/${PKGDIR:-"pkg"}
 
 software="sqlite-autoconf-${version:0:1}${version:2:2}0${version:5:1}00"
 URL="https://www.sqlite.org/2020/$software.tar.gz"
-[[ -d $software ]] || ( $WGET $URL; tar -xf $software.tar.gz )
+[[ -d $software ]] || ( $WGET --no-check-certificate $URL; tar -xf $software.tar.gz )
```
kgerheiser, line 39: I think we should avoid hardcoding this. The ideal solution is to have the OS fixed, but an alternative is to set `WGET=wget --no-check-certificate` in your config.

libs/build_tiff.sh
```diff
@@ -0,0 +1,49 @@
...
+if $MODULES; then
+    set +x
+    source $MODULESHOME/init/bash
+    module load hpc-$HPC_COMPILER
+    module list
+    set -x
+
+    prefix="${PREFIX:-"/opt/modules"}/$compiler/$name/$version"
+    if [[ -d $prefix ]]; then
+        [[ $OVERWRITE =~ [yYtT] ]] && ( echo "WARNING: $prefix EXISTS: OVERWRITING!";$SUDO rm -rf $prefix ) \
```
kgerheiser, line 23: Could you change this to:
```
      if [[ $OVERWRITE =~ [yYtT] ]]; then
          echo "WARNING: $prefix EXISTS: OVERWRITING!"
          $SUDO rm -rf $prefix
      else
          echo "WARNING: $prefix EXISTS, SKIPPING"
          exit 0
      fi
```

libs/build_tiff.sh
```diff
@@ -0,0 +1,49 @@
...
+
+cd ${HPC_STACK_ROOT}/${PKGDIR:-"pkg"}
+
+software=$name-$version
+URL="https://gitlab.com/lib${name}/lib${name}.git"
+[[ -d $software ]] || ( git clone $URL $software )
+[[ ${DOWNLOAD_ONLY} =~ [yYtT] ]] && exit 0
+[[ -d $software ]] && cd $software || ( echo "$software does not exist, ABORT!"; exit 1 )
+sourceDir=$PWD
+cmake -S . -B build -DCMAKE_INSTALL_PREFIX=$prefix
```
kgerheiser, line 42: Is there a reason to do an in-source build here, instead of creating a build directory? Out-of-source builds are preferrred.

kgerheiser (changes requested):

libs/build_proj.sh
aerorahul: perhaps an easier solution is to define
```
export WGET="wget --no-check-certificate
```
in your `config.sh`

natalie-perlin:
Will make a new and updated branch, and then will do another pull request

PR #386, opened 2022-02-15
Fix serial build of NetCDF when using pnetcdf

### LINKED ISSUE #382: NetCDF fails with enable_pnetcdf
**Describe the bug**
Compiling NetCDF fails with "...hpc-stack/libs/build_netcdf.sh: line 64: PNETCDF_ROOT: unbound variable". The lines containing "[[ $enable_pnetcdf =~ [yYtT] ]]" in libs/build_netcdf.sh should, I think, all check whether $mpi is defined so that the script won't try to use it when compiling serial NetCDF.

**To Reproduce**
-Clone repo
-Run build on, e.g., stack_custom.yaml
-Fails when attempting to build serial NetCDF

**System:**
What system(s) are you running the code on?
RedLine internal system

kgerheiser:
Fix #382

Check if `$mpi` is set before adding pnetcdf flags, so that serial build of NetCDF doesn't attempt


libs/build_netcdf.sh
```diff
@@ -61,11 +61,20 @@ curr_dir=$(pwd)
 
 LDFLAGS1="-L$HDF5_ROOT/lib"
 LDFLAGS2=$(cat $HDF5_ROOT/lib/libhdf5.settings | grep AM_LDFLAGS | cut -d: -f2)
-[[ $enable_pnetcdf =~ [yYtT] ]] && LDFLAGS4="-L$PNETCDF_ROOT/lib"
+if [[ $enable_pnetcdf =~ [yYtT] ]]; then
```
climbfuji, line 70: It seems like a contradiction to me if a user is not using MPI but requesting pnetcdf. Should the build process abort in the pnetcdf step?

kgerheiser: That's a good point, but hpc-stack always tries builds a serial + parallel build of NetCDF. In the serial case NetCDF should not try to build with MPI, but when it gets to the MPI build pnetcdf should be used.

This could be handled in build_stack.sh by unsetting pnetcdf before the serial and then restoring it.

aerorahul:
@kgerheiser Any update on this?

PR #394, opened 2022-02-24
Feature/linux centos intel

natalie-perlin:


libs/build_libtiff.sh
```diff
@@ -0,0 +1,56 @@
...
+    module load hpc-$HPC_COMPILER
+    module try-load cmake
+    module load zlib
+    module list
+    set -x
+
+    prefix="${PREFIX:-"/opt/modules"}/$compiler/$name/$version"
+    if [[ -d $prefix ]]; then
+        [[ $OVERWRITE =~ [yYtT] ]] && ( echo "WARNING: $prefix EXISTS: OVERWRITING!";$SUDO rm -rf $prefix ) \
+                                   || ( echo "WARNING: $prefix EXISTS, SKIPPING"; exit 1 )
```
kgerheiser: Could you change this to:

```
  if [[ -d $prefix ]]; then
      if [[ $OVERWRITE =~ [yYtT] ]]; then
          echo "WARNING: $prefix EXISTS: OVERWRITING!"
          $SUDO rm -rf $prefix
      else
          echo "WARNING: $prefix EXISTS, SKIPPING"
          exit 0
      fi
  fi
```

kgerheiser (changes requested):
Looks good. Could you just change the prefix overwrite block?

libs/build_libtiff.sh
natalie-perlin: updated build_libtiff.sh as requested

kgerheiser (approved):

kgerheiser:
@natalie-perlin could you update your branch. I'll merge this next.

PR #399, opened 2022-03-02
Fix NCO build on Hera

kgerheiser:
Disable ncap2 which requires Antlr libraries. This disables the C++ NCO libraries which we don't use.


Hang-Lei-NOAA (approved):

Hang-Lei-NOAA:
Tested and adopted on hera.

PR #404, opened 2022-03-03
Use ipolates=1 for wgrib2

kgerheiser:
wgrib2 2.0.7 does not support ip=3


Hang-Lei-NOAA (approved):
Agree with this as default

PR #413, opened 2022-03-23
Feature/mac gnu

natalie-perlin:
Added functionality for building on MacOS X (both M1/arm64 and Intel/x86_64)


stack/stack_mac.yaml
```diff
@@ -62,8 +62,8 @@ nccmp:
   version: 1.8.9.0
 
 nco:
-  build: YES
-  version: 5.0.6
+  build: NO
+  version: 4.9.3
```
kgerheiser: Why do you lower the version here? I've had success building with 5.0.6.

stack/stack_mac.yaml
```diff
@@ -277,7 +277,7 @@ pybind11:
 
 ecbuild:
   build: YES
-  version: 3.6.1
+  version: 2021.08.0
```
kgerheiser, line 281: Do you know what the difference is between the year version and the v3.x.x? I see they have a bunch of tags that kind of alternate between.

stack/stack_mac.yaml
```diff
@@ -322,24 +323,30 @@ yafyaml:
...
 proj:
-  build: NO
+  build: YES
   version: 7.1.0
   cmake_opts: "-DENABLE_CURL=OFF -DBUILD_PROJSYNC=OFF"
 
 miniconda3:
   build: NO
-  version: 4.7.12
+  version: 4.6.14
```
kgerheiser, line 349: What's the reason for lowering the version?

stack/stack_mac_gnu.yaml
```diff
@@ -0,0 +1,359 @@
+gnu:
```
kgerheiser, line 1: Can you say how this version differs specifically for GNU, compared to stack_mac?

stack/stack_mac_m1_gnu.yaml
```diff
@@ -0,0 +1,359 @@
+gnu:
```
kgerheiser, line 1: Same for this

kgerheiser (changes requested):

stack/stack_mac.yaml
natalie-perlin: Higher version could be left in the stack.yaml, as it is not being built on the MacOS. There is still an issue with the nco build on the MacOS

climbfuji: We've built 5.0.1 and 5.0.6 on macOS with clang+gfortran as well as gcc+gfortran.

stack/stack_mac.yaml
natalie-perlin: I could not yet find any information about the differences. The tag 2021.08.0 seems to correspond to 3.6.3 (maybe identical?); a VERSION file in 2021.08.0 has 3.6.3, and also these tags seem to be similar in github, at least from the first glance.

stack/stack_mac_gnu.yaml
natalie-perlin: I did not use stack_mac.yaml, maybe it could be reserved for using it with clang+gfortran. Yet to be tested with this combination.

stack/stack_mac_m1_gnu.yaml
natalie-perlin: This is only for GNU collection, when it is installed in advance of the build. MPI could be installed during the hpc-stack build.

stack/stack_mac.yaml
natalie-perlin: Tested building with nco v.5.0.6 - worked OK! Placing it back to the stack.

stack/stack_mac.yaml
natalie-perlin: Tested the software build for fms 3.6.3 -successful. This could be reverted to 3.x.x format.

stack/stack_mac_gnu.yaml
natalie-perlin: Update of the compiler suite to 11.2.x . Note that there is a bug in 11.1.0 compiler that prevents building of the w3nco library, v. 4.2.1. in particular. This compiler bug that affects the w3nco software is reported in 
[https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100755](url)
[https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100283](url)
And it is resolved in the GNU 11.2.x versions.

kgerheiser:
I resolved the conflict in your branch

natalie-perlin (commented):
All issues have been addressed.

It still says the merging is blocked. I will try updating a branch on NOAA-EPIC/hpc-stack:mac_gnu branch, then open pull requests from NOAA-EPIC to NOAA-EMC mac_gnu branch, and from mac_gnu to develop on NOAA-EMC side.
MergingBlocked

natalie-perlin:
All questions addressed, but some technical issues preventing merging ("merging blocked") within this pull request

kgerheiser:
I just had to re-approve the PR. Anyway #420 is open, so I'll use that.

natalie-perlin:
Reopened the PR, after updating of the feature/mac_gnu, it could be merged with the develop if all goes well:
 `NOAA-EMC:feature/mac_gnu` into `NOAA-EMC:develop`

stack/stack_mac.yaml
kgerheiser: What about the version of miniconda3 here? Why go from 4.7 to 4.6?

natalie-perlin: It could be updated to 4.7.12 ( not building )

kgerheiser: Nevermind, I this is fine. It's the version that `stack_noaa` uses, and it's not building.

kgerheiser (approved):

PR #415, opened 2022-03-24
Feature/cheyenne gnu

natalie-perlin:
Changes for Cheyenne, to build hpc-stack with modules restored/saved in module collection to avoid loading different versions of system modules


natalie-perlin:
need to PR to a different branch first

PR #417, opened 2022-03-25
Add option to build ESMF directly with tag and option for install_as

kgerheiser:
ESMF tagging scheme has changed recently, instead of being `ESMF_x_y_z_beta_snapshot_a_b`, it's `vx.y.zbab`

And set option in MAPL stack file with esmf_version to control which version it's built with.

For #407 so we don't have to edit the file manually to get the correct tag.

### DIFF
--- a/config/config_cheyenne_gnu.sh
+++ b/config/config_cheyenne_gnu.sh
@@ -43,5 +43,3 @@ export STACK_fms_FFLAGS="-march=core-avx2 -fallow-argument-mismatch"
 # Patch FMS
 export STACK_fms_PATCH="cheyenne_gnu_fms_mpp_util_mpi_inc.patch"
 
-export STACK_mapl_esmf_version="8_2_1_beta_snapshot_04"
-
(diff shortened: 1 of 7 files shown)


climbfuji (commented):
I like those changes. One thing that is missing - and that has always been missing in hpc-stack - is to be consistent w.r.t. debug/non-debug. The current setup is as follows:
- in PROD/REPRO mode, ESMF O2 and MAPL (optimized) are loaded; MAPL is built against ESMF O2
- in DEBUG mode, ESMF O0 and MAPL (optimized) are loaded; MAPL is built against ESMF02

I think we should be consistent and make sure that when ESMF debug is loaded, a version of MAPL that is built with ESMF debug is loaded, and if ESMF prod is loaded, a version of MAPL that is built with ESMF prod is loaded. We've done that in spack-stack already.

Hang-Lei-NOAA (approved):

kgerheiser:
@climbfuji we let one version of MAPL load either debug/non-debug ESMF. In the case of hpc-stack we have static libraries for MAPL and ESMF. The version of ESMF that MAPL is built with doesn't matter apart from the includes at compile time, which are the same because the ESMF code itself is identical. When you link the executable MAPL can link with the debug/non-debug ESMF depending on which module is loaded.

Issue #376, opened 2022-01-18
Inconsistent handling of prerequisite modules in lua modulefiles, some of which result in unintended behavior on some platforms


mkavulich:
**Describe the bug**

**The short summary: using `load(modulename)` in Lua modulefiles will not necessarily get you the module you expect. This is currently causing failures of the Short Range Weather Application on Cheyenne, and could theoretically result in unexpected behavior on any platform, up to and including build and runtime failures**

Many NCEPLIBS have dependencies on each other, this is obvious. Currently, however, there doesn't seem to be any uniformity in how these dependencies are handled; and some of the current methods are less than ideal. Here are the various methods used:

- ATLAS simply uses `prereq`, which causes the `module load` command to fail if the prerequisites are not already loaded. This seems like a natural way to handle required modules:

https://github.com/NOAA-EMC/hpc-stack/blob/44c8e077a21099fcc1883cd297783cdfa6c08b23/modulefiles/mpi/compilerName/compilerVersion/mpiName/mpiVersion/atlas/atlas.lua#L15-L18

- wgrib2 (in addition to others) uses a `load` command prior to the `prereq`. In addition to this being redundant (both `load` and `prereq` will fail if the module does not exist, and if it does exist, `prereq` will always succeed after a `load` of the same module), this has a dangerous side-effect of being able to overwrite the existing environment unexpectedly: if, for example, netcdf/4.7.4 is loaded from earlier in the hpc-stack process, but the system already has a built-in netcdf module, the `load(netcdf)` line will instead load that system default netcdf module, which may be a different version and incompatible with other hpc-stack modules!

https://github.com/NOAA-EMC/hpc-stack/blob/44c8e077a21099fcc1883cd297783cdfa6c08b23/modulefiles/mpi/compilerName/compilerVersion/mpiName/mpiVersion/wgrib2/wgrib2.lua#L15-L18

- MADIS (in addition to others) uses `always_load` in addition to `prereq`. This appears to be an even worse option to use, since in addition to the above-mentioned problems, `module unload` will attempt to **load** netCDF rather than unloading it; there doesn't seem to be any reason to do this.

https://github.com/NOAA-EMC/hpc-stack/blob/44c8e077a21099fcc1883cd297783cdfa6c08b23/modulefiles/mpi/compilerName/compilerVersion/mpiName/mpiVersion/madis/madis.lua#L15-L18

**To Reproduce**
On Cheyenne, run these commands and observe the environment changes:
```
module purge
[...]

PR #418, opened 2022-03-29
Change load, try_load, and prereq to depends_on

### LINKED ISSUE #410: Use depends_on instead of load/prereq in modules
Installing a new version of w3emc on WCOSS-Dell broke some application modules because of how `load()` chooses the default version. We have a few instances of `try_load()` as well, but that doesn't do what we think it does. It is just identical to `load()` except that it doesn't throw an error if the module is not loaded. `depends_on()` is the correct function, which only loads the module if it's not already loaded.

The LMod documentation has this to say about `depends_on()`

```
This choice loads module “A” on the users behalf if it not already loaded.
```

https://lmod.readthedocs.io/en/latest/098_dependent_modules.html

kgerheiser:
depends_on is the correct behavior in most circumstances. If the dependent module is already loaded it will use that, but if not will try to load it. This way users can mix and match their own versions without LMod changing their selection and forcing the default.

This has been a recurring problem where module load order matters forcing users to override previously loaded modules through `load`.

depends_on combines the functionality of load/preqreq in a single step.

In some cases such as MAPL, compiler/MPI meta modules, and HDF5 with szip load/try_load where the module has to exist or might not exist is the correct behavior.

And not sure why we had `always_load` in some cases.

See https://lmod.readthedocs.io/en/latest/098_dependent_modules.html

Fix #410 and #376

### DIFF
--- a/modulefiles/compiler/compilerName/compilerVersion/esmf/esmf.lua
+++ b/modulefiles/compiler/compilerName/compilerVersion/esmf/esmf.lua
@@ -11,10 +11,7 @@ local compNameVerD = compNameVer:gsub("/","-")
 
 conflict(pkgName)
 
-always_load("hdf5")
-always_load("netcdf")
-prereq("hdf5")
-prereq("netcdf")
+depends_on("netcdf")
 
 local opt = os.getenv("HPC_OPT") or os.getenv("OPT") or "/opt/modules"
 
--- a/modulefiles/compiler/compilerName/compilerVersion/impi/impi.lua
+++ b/modulefiles/compiler/compilerName/compilerVersion/impi/impi.lua
@@ -17,11 +17,6 @@ family("mpi")
 conflict(pkgName)
 conflict("mpich","openmpi")
 
-always_load("intel/17.0.1")
-prereq("intel/17.0.1")
-
-try_load("szip")
-
 local opt = os.getenv("HPC_OPT") or os.getenv("OPT") or "/opt/modules"
 local base = "/opt/intel17/compilers_and_libraries_2017.1.132"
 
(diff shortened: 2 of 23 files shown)


modulefiles/compiler/compilerName/compilerVersion/impi/impi.lua
```diff
@@ -17,11 +17,6 @@ family("mpi")
 conflict(pkgName)
 conflict("mpich","openmpi")
 
-always_load("intel/17.0.1")
-prereq("intel/17.0.1")
-
-try_load("szip")
```
kgerheiser, line 23: No reason for impi to load szip.

modulefiles/compiler/compilerName/compilerVersion/esmf/esmf.lua
```diff
@@ -11,10 +11,7 @@ local compNameVerD = compNameVer:gsub("/","-")
 
 conflict(pkgName)
 
-always_load("hdf5")
-always_load("netcdf")
-prereq("hdf5")
-prereq("netcdf")
+depends_on("netcdf")
```
kgerheiser, line 14: NetCDF will load HDF5 transparently.

modulefiles/compiler/compilerName/compilerVersion/openmpi/openmpi.lua
```diff
@@ -14,8 +14,6 @@ family("mpi")
 conflict(pkgName)
 conflict("mpich","impi")
 
-try_load("szip")
```
kgerheiser, line 17: OpenMPI does not need szip

modulefiles/core/hpc-cray-intel/hpc-cray-intel.lua
```diff
@@ -13,7 +13,6 @@ conflict("hpc-gnu", "hpc-gcc")
 local compiler = pathJoin("intel",pkgVersion)
 load(compiler)
 prereq(compiler)
-try_load("mkl")
```
kgerheiser, line 16: Compiler should not load MKL

aerorahul (approved):
changes look good.

climbfuji (approved):

Hang-Lei-NOAA (approved):
A great change to enable larger compatibility.

PR #428, opened 2022-04-12
Prevent doc and README changes from triggering Github Action workflow

kgerheiser:
And add manual dispatch to rest of actions

### DIFF
--- a/.github/workflows/build_intel.yaml
+++ b/.github/workflows/build_intel.yaml
@@ -1,6 +1,10 @@
 name: Build Intel
 on:
   pull_request:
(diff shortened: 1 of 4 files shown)


Hang-Lei-NOAA (approved):

kgerheiser:
Unfortunately, this doesn't work the way I had hoped when using the branch protection rules. The workflow will not run and the checks will never complete, and there will be no way of merging.

kgerheiser:
Closing until I can find a workaround to skip the status checks when certain files change.

PR #429, opened 2022-04-12
Make commands portable to older macs

gspetro-NOAA:
I changed two lines of code so that the commands would work on older macs running an older bash version. 

@natalie-perlin


kgerheiser:
Can you update your branch to be even with develop?

gspetro-NOAA:
@kgerheiser I thought I had... When I look at the diff on Git, it's just showing the two lines that I changed. Everything else is the same. Are you seeing something else?

kgerheiser (approved):

PR #444, opened 2022-04-29
Feature/cheyenne

natalie-perlin:
[Changes for Cheyenne to build with modules saved/restored in Lmod module collections

Module collections are saved in /glade/u/home/nperlin/.lmod.d/ directory on Cheyenne


config/config_cheyenne_gnu.sh
```diff
@@ -16,30 +16,23 @@ export MAKE_VERBOSE=N
...
 export STACK_pnetcdf_FFLAGS="-fallow-argument-mismatch -fallow-invalid-boz"
 export STACK_madis_FFLAGS="-fallow-argument-mismatch -fallow-invalid-boz"
+export STACK_fms_FFLAGS="-march=core-avx2 -fallow-argument-mismatch"
 
 # Build FMS with AVX2 flags
 export STACK_fms_CFLAGS="-march=core-avx2"
-export STACK_fms_FFLAGS="-march=core-avx2 -fallow-argument-mismatch"
-
-# Patch FMS
-export STACK_fms_PATCH="cheyenne_gnu_fms_mpp_util_mpi_inc.patch"
```
climbfuji, line 44: Is this no longer needed with mpt-2.25? We just recently went through this exercise and found that it was still needed for the ufs-weather-model (with mpt-2.22 and gnu 10.1.0 at least).

libs/build_hdf5.sh
```diff
@@ -85,3 +87,8 @@ VERBOSE=$MAKE_VERBOSE make -j${NTHREADS:-4}
 [[ -z $mpi ]] && modpath=compiler || modpath=mpi
 $MODULES && update_modules $modpath $name $version
 echo $name $version $URL >> ${HPC_STACK_ROOT}/hpc-stack-contents.log
+# save the current modules list to the hpc-hdf5 collection
```
climbfuji, line 90: I don't see how this fixes the problem with the CISL modules always taking precedence over modules added by users. Can you explain, please?

climbfuji (commented):
How would one use those stored module collections? If they are stored in a personal directory, then that is a no-go for any release. I would also like to understand how this helps with the problem that if the same version of a software is available from the CISL modules, it will take precedence.

libs/build_hdf5.sh
natalie-perlin: @climbfuji - storing module collections is a Lmod functionality, it uses `module save <lmod_collection>`   and `module restore <lmod_collection>` commands, and the collections of modules are saved in user's home directory `~/.lmod.d/`.  This is a standard way similar to user space configuration using ~/.ssh/ directories, or ~/.matlab/ , ~/.vim, ~/.gitconfig .
An example of such a collection, the file `~/.lmod.d/hpc-mpi-esmf.ch`  on cheyenne in my home directory, is attached, except with the name modified to the allowed format to attach it, with the added *.TXT extension.

Additional functionality is used to prevent the system to load default module versions, when a specific version of the same software module is needed and it is already loaded. The syntax uses 'module is-loaded' construct. A couple of  examples:
`module is-loaded hdf5 || module load hdf5`
`module is-loaded cmake || module try-load cmake`
`module is-loaded netcdf || module load netcdf`


[hpc-mpi-esmf.ch.TXT](https://github.com/NOAA-EMC/hpc-stack/files/8637652/hpc-mpi-esmf.ch.TXT)

natalie-perlin: I'm still cleaning up the code a little more, to separate the builds with higher/progressive module versions on Cheyenne and the added functionality with the module collections, and will update this PR.

config/config_cheyenne_gnu.sh
natalie-perlin: The code to be updated, to leave only the functionality of Lmod module collections, and  to avoid use of different modules from those in the current "develop" branch.

natalie-perlin:
Adding log files from cheyenne's build that used "module restore"  from Lmod collections.
Built with gnu/11.2.0, mpi/2.25.
HPC-stack contents: **hpc-stack-contents_cheyenne_gnu11.2.log**
Log file in the building directory: **build_stack_cheyenne_gnu11.2.log**
Compressed archived log/ directory with individual log files: **log_cheyenne_gnu11.2.tgz** (please note that it has been renamed to log_cheyenne_gnu11.2.tgz.zip to comply with naming formats allowed to be attached to the comment). 
Notifications on modules being restored and module list following a "module restore" command could be found in log files, by searching strings **"Restoring modules "*
[hpc-stack-contents_cheyenne_gnu11.2.log](https://github.com/NOAA-EMC/hpc-stack/files/8701761/hpc-stack-contents_cheyenne_gnu11.2.log)
[build_stack_cheyenne_gnu11.2.log](https://github.com/NOAA-EMC/hpc-stack/files/8701770/build_stack_cheyenne_gnu11.2.log)
[log_cheyenne_gnu11.2.tgz.zip](https://github.com/NOAA-EMC/hpc-stack/files/8701813/log_cheyenne_gnu11.2.tgz.zip)
*

natalie-perlin:
Closing the PR in its current form, it's behind other updates.

PR #445, opened 2022-04-29
Bugfix/wgrib2 ubuntu build

natalie-perlin:
Changing the defaults in makefile _not_to_build_ libraries NETCDF3, PNG, JASPER. Libraries NETCDF3 and JASPER are very old, and contain config.guess file with the timestamp 
`config.guess timestamp = 2006-07-02` ( ./grib2/jasper-1.900.1/acaux/config.guess) that result in configure error when checking build system type. This error appears when building on Ubuntu 20.04, aarch64 (arm64) architecture.

JASPER,NETCDF, and PNG libraries are also built separately for the HPC-stack, so there is no need for them to be build additionally as older versions while compiling wgrib2


kgerheiser:
@natalie-perlin this branch contains a bunch of other changes. Can you cleanup this PR so that it's based off of develop and has just what's in the title?

natalie-perlin:
Updated, changing the only file that fixes the error described

kgerheiser:
Thank you. 

So, wgrib2 is weird and it packages and builds its own dependencies with no knowledge of the outside NetCDF/Jasper. By disabling these the functionality is also lost.

It can either be defaulted to off and enabled specifically where it's needed, or leave the default and disable it in the stack file, like:

```
wgrib2:
  build: YES
  netcdf3: NO
```

Not positive about Jasper, but I do know that the NetCDF option is used by some users.

climbfuji:
What's the status of this PR?

aerorahul:
The changes in this PR will render wgrib2 defunct as @kgerheiser points out. 
wgrib2 builds with the libraries it packs in its release. Attempting to use system installed libraries often results in a broken wgrib2 installation.

kgerheiser:
The Jasper/NetCDF functionality of wgrib2 is not widely used AFAIK. I think it would be fine to disable it by default and leave it enabled in `stack_noaa.yaml` because it causes all these build issues.

natalie-perlin:
outdated

PR #446, opened 2022-05-06
Minor documentation updates for installation on MacOS

gspetro-NOAA:
- Added installation instructions for openssl@3 package
- Modified Lmod installation syntax to reflect more common usage
- Changed EMC_post reference to UPP
- Fixed typos

@natalieperlin


kgerheiser (approved):

docs/source/mac-install.rst
```diff
@@ -207,13 +232,15 @@ For Option 1:
...
+   export BASH_ENV="/opt/homebrew/opt/lmod/init/profile"
+   source $BASH_ENV
 
 For Option 2:
 
 .. code-block:: console 
 
-    source /usr/local/opt/lmod/init/profile
+   export BASH_ENV="/usr/local/opt/lmod/init/profile"
+   source $BASH_ENV
```
kgerheiser, line 243: What's the point of making it a Bash variable before sourcing?

gspetro-NOAA: I think @natalie-perlin was just switching it to a more common (albeit less concise) way of doing this, but I'll let her weigh in. I presume it would also allow users to use this variable again later, if desired, although I don't know under what circumstances they would need to.

natalie-perlin: This syntax is similar to the standard Lmod initialization practice for the shell (bash, as suggested to be used by the supporting documentation).

natalie-perlin: BASH_ENV is also a standard environmental variable used in bash shells

PR #461, opened 2022-06-03
HPC-stack documentation updates with the example for Linux Ubuntu20.04

natalie-perlin:
Documentation updates only, with the example for Linux Ubuntu 20.04.


kgerheiser (approved):

kgerheiser:
Looks good. I just used similar instructions when setting up a new Ubuntu VM.

PR #474, opened 2022-06-29
Building UPP bugfix.

t-brown:
When building UPP without modules it fails as it is unable
to find the dependent FORTRAN module files.


kgerheiser:
What version of UPP are you building? These variables are not used by the CMake build.

t-brown:
I was building `stack/stack_gfsv16.2.yaml`, which looks like it is pulling in `upp_v8.1.0`.

kgerheiser (approved):

PR #476, opened 2022-07-05
Feature/conda macos linux

natalie-perlin:
##  SUMMARY OF MODIFICATIONS

Builds for linux_gnu and macos_gnu (both Intel and M1 chips). GNU gcc compiler is preinstalled, openmpi v4.2.1 is build as a part of the HPC-stack. Miniconda3 is build, and regional_workflow environment is created  under the miniconda3 module. Miniconda3 latert version is checked out.

ESMF v8.3.0b09, builds well with mapl 2.12.2  (Build of mapl-2.11.0 fails with the ESMF v8.3.0b09).

## FILES MODIFIED
config/config_linux_gnu.sh
config/config_macos_gnu.sh
stack/stack_macos_gnu.yaml
stack/stack_linux_gnu.yaml
libs/build_mapl.sh
libs/build_miniconda3.sh
libs/build_nceplibs.sh


kgerheiser:
This looks mostly alright, but why are there so many `stack_` files? Does there need to be one for every machine? It makes sense in some cases, but why does there need to be `stack_jet`, `stack_orion`, `stack_hera`?

build_stack.sh
```diff
@@ -188,12 +188,12 @@ build_lib sigio
 build_lib sfcio
 build_lib gfsio
 build_lib w3nco
+build_lib nemsio
```
kgerheiser, line 191: This needs to come after w3emc.

kgerheiser (changes requested):

natalie-perlin:
stack_jet, stack_orion, stack_hera - GNU versions indeed could be removed, if there is no intent to keep two versions of compilers, `intel `and `gnu`. The GNU versions of stack_...yaml files were created as an attempt to test different compilers.

build_stack.sh
natalie-perlin: Some of the changes were to build  the hpc-stack with w3emc/2.7.3 as needed for the SRW release/public-v2. The "develop" branch has an update w3emc/2.9.2. 
Will revise the changes and break them into smaller PRs.

natalie-perlin:
Closing the PR, and will break down the changes into more specific ones into different branches ("develop", "release/srw-public-v2".

PR #479, opened 2022-07-13
build option and configuration for linux_gnu, building miniconda3 module and regional_workflow virtual environment

natalie-perlin:
## SUMMARY OF CHANGES (updated 15-July-2022)
 
- Configuration and build options for general Linux with GNU compilers. 
- Turn off building documentation for libtiff, which relies on sphinx-build. It was reported to get an error during hpc-stack build, when sphinx-build is found and the documentation for libtiff is attempted to be built.

- Updated `miniconda3` to build a base miniconda3 module
- Create a virtual environment `regional_workflow` with the packages needed for the SRW, when it is requested in stack_....yaml. Use an adapted `./libs/build_condaenv.sh` script. It also builds the module with the same name, `regional_worflow`. Virtual environment could then be activated/deactivated in two ways:
a) Load miniconda3 module, and then by using "conda activate regional_workflow" , "conda deactivate"
b) Load hpc-${HPC_PYTHON} module, miniconda3 module, and then use "module load regional_workflow" , "module unload regional_workflow".

- When building MAPL, use the STACK_FFLAGS variable if defined

## FILES CHANGED

./config/config_linux_gnu.sh
./stack/stack_linux_gnu.yaml
./libs/build_libtiff.sh
./libs/build_miniconda3.sh
./libs/build_condaenv.sh
./build_stack.sh
./pyvenv/regional_workflow.yml
./pyvenv/regional_workflow.txt
./libs/build_mapl.sh


libs/build_miniconda3.sh
```diff
@@ -67,7 +67,13 @@ source $prefix/etc/profile.d/conda.sh
 echo "disabling conda auto updates"
 conda config --system --set auto_update_conda False
 echo "install $version of conda"
-conda install -yq conda=$version
+#conda install -yq conda=$version
+conda config --add channels conda-forge
+conda config --set channel_priority strict
+conda create -n regional_workflow -y jinja2 pyyaml f90nml
```
aerorahul: I don't think this is acceptable in the stack source code.
It appears that you are creating a conda env called `regional_workflow`.
That should be a recipe.

aerorahul: See the other examples on how conda or virtual envs are created for e.g. `ufswm`

edwardhartnett:
@natalie-perlin are you aware we will soon be transitioning to spack-stack, and hpc-stack will be deprecated? See https://github.com/NOAA-EMC/spack-stack

libs/build_miniconda3.sh
natalie-perlin: Thanks for the note. Suggesting an updated version of the build_miniconda3.sh. 
Yaml file stack_linux_gnu.yaml contains the following:
miniconda3:
  build: YES
  version: 4.12.0
  pyversion: py39
  pyvenv: regional_workflow
  
 where `pyvenv` is the name of the recipe file, located under ./hpc-stack/pyvenv/, e.g., `regional_workflow.yml`.
 A requested virtual environment with the same name as `$pyvenv` is created only if this variable is set in stack...yaml file for miniconda3, and only if the requirements `$pyvenv.yml` file exists.

natalie-perlin: Updating the description of the PR and the changes introduced with the recent update.

natalie-perlin:
@edwardhartnett  - yes, absolutely! Planning to move to the spack-stack asap. 
In the documentation for the recent SRW release (v2.0.0), however, only the hpc-stack build recipe was suggested for the users. So if the support for the SRW release v2.0.0 is considered, the hpc-stack recipe needs to be up-to-date. There was a recent help request at the UFS forums for the build of the SRW on linux_gnu platform, thus the particular update of the branch.

https://forums.ufscommunity.org/threads/compilation-error-symbol-rmissing-not-found-module-kinds

libs/build_miniconda3.sh
```diff
@@ -8,6 +8,8 @@ set -eux
 name="miniconda3"
 version=${2:-${STACK_miniconda3_version:-"latest"}}
 pyversion=${3:-${STACK_miniconda3_pyversion:-}}
+pyvenv=${4:-${STACK_miniconda3_pyvenv:-}}
```
aerorahul: The job of this script `build_miniconda3.sh` is to build and install miniconda3 and not a conda env.
To install a conda env., one would install miniconda3 (this script), and then use the `libs/build_condaenv.sh` to build the conda env. using the miniconda3 installed by this script.

See examples in `develop` and instructions [here](https://hpc-stack.readthedocs.io/en/latest/hpc-notes.html#adding-a-new-library-or-package)

natalie-perlin: Updated a recipe to build a regional_workflow conda environment, when it is requested in `./stack/stack...yaml` file. It uses an updated `./libs/build_condaenv.sh` script; no need to create an additional `build_...sh` script.

libs/build_condaenv.sh
```diff
@@ -89,5 +93,5 @@ set -x
 set -u
 
 # generate modulefile from template
-$MODULES && update_modules python $name $version $python_version
+# $MODULES && update_modules python $name $version $python_version
```
kgerheiser: Why is module generation disabled?

natalie-perlin: There is no module to create. The script build_condaenv.sh only adds a conda virtual environment `regional_workflow` to the existing miniconda3 module.

aerorahul: How do you activate the environment without calling conda calls?
The module will allow you to do:
`module load regional_workflow`
ps. I am happy to explain offline.

natalie-perlin: Updated a comment before the line with the update_modules command

libs/build_condaenv.sh
```diff
@@ -21,22 +21,30 @@ rqmts_file=${HPC_STACK_ROOT}/pyvenv/$rqmts
...
   set +x
   source $MODULESHOME/init/bash
-  module load hpc-$HPC_PYTHON
+  module load hpc
+  if [[ -n "${STACK_miniconda3_version:-} " ]]; then
+    module is-loaded python  && module unload python
+    module load miniconda3/${STACK_miniconda3_version:-}
+  else
+    module load hpc-$HPC_PYTHON
+  fi
```
aerorahul: `hpc-$HPC_PYTHON` would be `hpc-miniconda3` if `PYTHON=miniconda3`, and it will load `miniconda3`
This section is incorrect.

natalie-perlin: This is a bit tricky. When `miniconda3` is requested to be being built as a part of the hpc-stack (and is not used as as system module) , `$HPC-python` variable needs needs to be set as `hpc-python` not `hpc-miniconda3` , which is done in config file. This is needed for the other modules to be built prior to `miniconda3` in the build_stack.sh. 

Variable `$STACK_miniconda3_version` is not empty only when the miniconda3 is built as a part of the stack. The module `miniconda3/${STACK_miniconda_version}` could only be loaded after any existing python module (if a system-wide module exists) is unloaded.

Any ideas?

natalie-perlin: If miniconda3 is build locally and not as a part of `hpc-stack`, the module loaded will be `hpc-miniconda3`. So there seems to be no conflict. I could only suggest to use $SUDO to create a conda environment, in case the system-wide installation of miniconda3 exist:

$SUDO conda env create --file $rqmts_file

(this is needed for the user to write into the $CONDA_ENVS_PATH/ directory which could be a conda directory outside the user's stack installation.)

libs/build_condaenv.sh
natalie-perlin: @aerorahul - As of current development of the he SRW App, it uses a separate module `./ufs-srweather-app/modulefiles/wflow_<platform>` to load the miniconda3, if the conda environment is used, and to activate a pre-configured environment, similarly to the following:

```
module load miniconda3/
conda activate regional_workflow
```
 If the python environment is used for the SRW workflow, this `wflow_<platform>` module would contain something like `source $VENV/bin/activate`, where the $VENV is the path of the `regional_workflow` venv directory. But I think there is a plan to move towards using `conda` for the SRW.
 
 So the current hpc-stack recipe for `regional_workflow` only adds a virtual conda environment, not a separate module. The SRW app does not expect to have or to load an additional module `regional_workflow`, only the environment to be activated/deactivated.

aerorahul: you could easily replace 
`conda activate regional_workflow`
with
`module load regional_workflow`.

When using the python environment, the module would still be `module load regional_workflow`.  
So you see, it matters not whether you build `regional_workflow` via conda or python virtual environment for the user, the user only needs to load the module for `regional_workflow`

The `module load regional_workflow` will activate and `module unload regional_workflow` will deactivate.
The user does not need to worry about native conda or venv calls.

aerorahul: I will also note that the `module load` is the supported mechanism and not `conda activate` or `source $VENV/bin/activate`. 
This allows one to not know the inner details of how the module was built.

libs/build_condaenv.sh
natalie-perlin: @aerorahul @kgerheiser -
I addressed the issue by creating both virtual environment and the module with the same name. The summary of changes on the top of the PR description  now reads:

- Create a virtual environment regional_workflow with the packages needed for the SRW, when it is requested in stack_....yaml. Use and adapted ./libs/build_condaenv.sh script. It also builds the module with the same name regional_worflow. Virtual environment could then be activated/deactivated in two ways:
a) Load miniconda3 module, and then by using "conda activate regional_workflow" , "conda deactivate"
b) Load hpc-${HPC_PYTHON} module, miniconda3 module, and then use "module load regional_workflow" , "module unload regional_workflow".

natalie-perlin: Please see attached a screenshot from the build on Orion, showing the two ways the conda environment is activated
conda_hpc-python_regional_workflow

libs/build_condaenv.sh
```diff
@@ -26,6 +26,10 @@ if $MODULES; then
   set +x
   source $MODULESHOME/init/bash
   module load hpc-$HPC_PYTHON
+  if [[ -n "${STACK_miniconda3_version:-} " ]]; then
+    module is-loaded python  && module unload python
+    module load miniconda3/${STACK_miniconda3_version:-}
+  fi
```
aerorahul, line 32: You shouldn't need to do this between lines 29-32.

The correct way would be to define `HPC_PYTHON=miniconda3/miniconda3_version` in your `config.sh` file.  
When `setup_modules.sh` is executed, it will create a `hpc-miniconda3` modulefile.
Then L28, `module load hpc-$HPC_PYTHON` will load the correct `miniconda3/miniconda3_version` module.

aerorahul:
@natalie-perlin 
Is the still relevant?

PR #495, opened 2022-10-15
NCO compliance for third-party software on FISMA High Systems

aerorahul:
During the review of software requests for installation on NOAA's operational HPC WCOSS2, NCO IT team flagged several software as non-compliant and were rejected.

This PR:
- addresses the NCO IT security team reviews and concerns.
- addresses the concerns for `ecbuild`, `eckit`, `fckit`, `atlas` and `yafyaml`.
- repoints EMCWF softwares to ECMWF authoritative repositories as default rather than JCSDA.

The IT security manual is available within the NOAA domain.

### DIFF
--- a/libs/build_ecbuild.sh
+++ b/libs/build_ecbuild.sh
@@ -3,8 +3,8 @@
 set -eux
 
 name="ecbuild"
-repo=${1:-${STACK_ecbuild_repo:-"jcsda"}}
-version=${2:-${STACK_ecbuild_version:-"release-stable"}}
+repo=${1:-${STACK_ecbuild_repo:-"ecmwf"}}
+version=${2:-${STACK_ecbuild_version:-"master"}}
 
 if $MODULES; then
   set +x
@@ -18,7 +18,7 @@ if $MODULES; then
       if [[ $OVERWRITE =~ [yYtT] ]]; then
           echo "WARNING: $prefix EXISTS: OVERWRITING!"
           $SUDO rm -rf $prefix
-          $SUDO mkdir $prefix 
+          $SUDO mkdir $prefix
       else
           echo "WARNING: $prefix EXISTS, SKIPPING"
           exit 0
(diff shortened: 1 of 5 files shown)


Hang-Lei-NOAA (approved):

aerorahul:
@climbfuji 
I am going to merge this as it is required at WCOSS2.
I am interested in your thoughts on how to adopt this to spack-stack.

libs/build_ecbuild.sh
```diff
@@ -35,11 +35,20 @@ URL="https://github.com/$repo/$name.git"
...
+  sed -r -i 's|http[]://[a-zA-Z0-9./-]*/test-data|DISABLED_BY_DEFAULT|g' cmake/ecbuild_check_urls.cmake
+  sed -r -i 's|http[]://[a-zA-Z0-9./-]*/test-data|DISABLED_BY_DEFAULT|g' cmake/ecbuild_get_test_data.cmake
+fi
+
 [[ ${DOWNLOAD_ONLY} =~ [yYtT] ]] && exit 0
 [[ -d build ]] && $SUDO rm -rf build
 mkdir -p build && cd build
 
-cmake -DCMAKE_INSTALL_PREFIX=$prefix ..
+../bin/ecbuild --prefix=$prefix ..
```
climbfuji, line 51: Is there a particular reason why `ecbuild` is used here, and not `cmake`?

libs/build_ecbuild.sh
```diff
@@ -35,11 +35,20 @@ URL="https://github.com/$repo/$name.git"
 [[ -d $software ]] && cd $software || ( echo "$software does not exist, ABORT!"; exit 1 )
 
 git checkout $version
+
+# Conform to NCO IT FISMA High Standards
+if [[ ${NCO_IT_CONFORMING:-"NO"} =~ [yYtT] ]]; then
+  sed -r -i 's|ssh://[a-zA-Z0-9.@]*||g' cmake/compat/ecmwf_git.cmake
+  sed -r -i 's|http[]://[a-zA-Z0-9@${}_./]*||g' cmake/compat/ecmwf_git.cmake
+  sed -r -i 's|http[]://[a-zA-Z0-9./-]*/test-data|DISABLED_BY_DEFAULT|g' cmake/ecbuild_check_urls.cmake
```
climbfuji, line 43: I wonder if `ecbuild` has a macro (when building `ecbuild` itself) to do what lines 43 and 44 are doing.

climbfuji (commented):
This all looks reasonable and is completely doable in `spack-stack`. I suggest we define a variant `+fismahigh` or `+ncofisma` or `+fisma` (or whatever better suggestion you may have), which is by default turned off, but if turned on applies these patches after the installation.

Note that we are using the authoritative ecmwf repositories in spack-stack, not the jcsda repositories.

libs/build_ecbuild.sh
aerorahul: Just following the installation directions provided by the developers.
https://github.com/ecmwf/ecbuild/blob/develop/INSTALL.rst

climbfuji: Ok, fair enough. I think that spack-stack always builds with plain cmake, but as long as the ecbuild include/macro loads are up first it works the same way.

libs/build_ecbuild.sh
aerorahul: Lines 41 and 42 define `ECMWF_SSH` and `ECMWF_HTTPS` addresses to ECMWF internal git servers.  No one at NOAA (or external users) will ever have that.  Any attempt to access them will basically result in a timeout.  
I think these lines are for convenience in applications using `ecbuild` at ECMWF so users don't have to memorize them.  
Line 43 is benign, as one can provide the URL at the command-line to over-ride the `DISABLE_BY_DEFAULT`.
Removing that capability is far more intrusive.

PR #520, opened 2023-04-26
update checkout version and clean after success

aerorahul:
This PR:
- updates the checkout action to v3 to remove annoying Github Actions warnings
- cleans out the `pkg/` directory after a successful action.  This should not be necessary as the scratch space is cleaned out automatically, but it does no harm.


natalie-perlin (approved):
Thank you for the fixes!

